PR #6, opened 2023-08-25
Add support for projected volume kubeconfig

timuthy:

/area security
/kind enhancement

This PR adds the option to configure a projected volume for the validator which can be used as a kubeconfig. It is for example needed, if operators generate their kubeconfigs for the virtual garden cluster via the [TokenRequestor](https://github.com/gardener/gardener/blob/master/docs/concepts/resource-manager.md#tokeninvalidator-controller).

**Release note**:

```feature operator
The `gardener-extension-shoot-rsyslog-relp-admission` chart allows to optionally configure a projected volume based kubeconfig.
```

### DIFF
--- a/charts/gardener-extension-shoot-rsyslog-relp-admission/charts/runtime/templates/deployment.yaml
+++ b/charts/gardener-extension-shoot-rsyslog-relp-admission/charts/runtime/templates/deployment.yaml
@@ -47,6 +47,9 @@ spec:
         {{- if .Values.kubeconfig }}
         - --kubeconfig=/etc/gardener-extension-shoot-rsyslog-relp-admission/kubeconfig/kubeconfig
         {{- end }}
+        {{- if .Values.projectedKubeconfig }}
+        - --kubeconfig={{ required ".Values.projectedKubeconfig.baseMountPath is required" .Values.projectedKubeconfig.baseMountPath }}/kubeconfig
+        {{- end }}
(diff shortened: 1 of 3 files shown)


charts/gardener-extension-shoot-rsyslog-relp-admission/charts/runtime/templates/deployment.yaml
```diff
@@ -47,6 +47,9 @@ spec:
         {{- if .Values.kubeconfig }}
         - --kubeconfig=/etc/gardener-extension-shoot-rsyslog-relp-admission/kubeconfig/kubeconfig
         {{- end }}
+        {{- if .Values.projectedKubeconfig }}
```
plkokanov, line 50: Shouldn't this be in an `else if` block or do we expect to never have both `.Values.kubeconfig` and `.Values.projectedKubeconfig` defined?

plkokanov (commented):
Thanks for the PR. I just have one question.

charts/gardener-extension-shoot-rsyslog-relp-admission/charts/runtime/templates/deployment.yaml
timuthy: Yes, we expect that both it not set at the same time since it doesn't make much sense.

plkokanov (approved):
/lgtm

PR #11, opened 2023-09-01
Fix e2e test flake for shoots created with the `shoot-rsyslog-relp` extension already enabled

plkokanov:
This PR fixes another issue that could cause flakes when a shoot is created with the `shoot-rsyslog-relp` extension already enabled. After rsyslog-relp is installed on the shoot node, we should wait for at least 15 seconds to make sure that the script which configures rsyslog on the node has a chance to run.

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/test/e2e/create_enabled_hibernate_reconcile_delete_test.go
+++ b/test/e2e/create_enabled_hibernate_reconcile_delete_test.go
@@ -41,7 +41,7 @@ var _ = Describe("Shoot Rsyslog Relp Extension Tests", func() {
 		defer cancel()
 		verifier, err := newVerifier(ctx, f.Logger, f.ShootFramework.SeedClient, f.ShootFramework.ShootSeedNamespace(), "local", f.Shoot.Name, string(f.Shoot.UID))
 		Expect(err).NotTo(HaveOccurred())
-		verifier.verifyThatLogsAreSentToEchoServer(ctx, "test-program", "1", "this should get sent to echo server")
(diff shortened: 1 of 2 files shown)


test/e2e/create_enabled_hibernate_reconcile_delete_test.go
```diff
@@ -36,6 +36,9 @@ var _ = Describe("Shoot Rsyslog Relp Extension Tests", func() {
 		By("Install rsyslog-relp unit on Shoot nodes")
 		Expect(execInShootNode(ctx, f.ShootFramework.SeedClient, f.Logger, f.ShootFramework.ShootSeedNamespace(), "apt-get update && apt-get install -y rsyslog-relp")).To(Succeed())
 
+		By("Wait for rsyslog to get configured on Shoot node")
+		time.Sleep(15 * time.Second)
```
ialidzhikov, line 40: You could use `Eventually` and put the verifications in that `Eventually` with the given timeout (15s) instead of sleeping.

plkokanov: The verification already use an `Eventually` or `Consistently` blocks depending on what they check , perhaps the better solution here would have been to simply increase the timeouts of those blocks to 15s. I'll try it out.

ialidzhikov: Yep, sorry, that I missed this. Sounds better to increase the corresponding `Eventually` and `Consistently` timeouts. 👍

ialidzhikov (approved):
/lgtm

PR #13, opened 2023-09-04
Add objectSelector to validating webhook config for shoots

### LINKED ISSUE #12: Make the shoot validaitng admission webhook only validate `Shoots` which enable the `shoot-rsyslog-relp` extension
/kind enhancement

Add an `objectSelector` to the `shoot-rsyslog-relp` `ValidatingAdmissionWebhook` which is called to validate `Shoot`s. The `objectSelector` should ensure that the webhook is only called for `Shoot`s that enable the `shoot-rsyslog-relp` extension.
The following can serve as an example: https://github.com/gardener/gardener-extension-registry-cache/blob/5e2ed0ef195c054cd81c04f5b331cf2c92bdf4a0/charts/admission/charts/application/templates/validatingwebhook.yaml#L19-L23

**Why is this needed**:
Currently the `ValidatingAdmissionWebhook` which is called when validating `Shoot` resources to check if the `shoot-rsyslog-extension` is configured is called for all `Shoot`s. To prevent issues and ease the work of the extension the validation should only be done of the `shoot-rsyslog-extension` spec is actually present in the `Shoot` resource

plkokanov:
/kind enhancement

This PR adds an `objectSelecor` to the `ValidatingWebhookConfiguration` which validates whether the `shoot-rsyslog-relp` extension is configured properly on `Shoot` resources. With the new `objectSelector` only shoots that actually enable the `shoot-rsyslog-relp` extension are considered for validation. 

This makes use of the fact that the gardener `ExtensionLabels`  admission plugin annotates the `Shoot` with the `extensions.extensions.gardener.cloud/shoot-rsyslog-relp: 'true'` label when the extension is enabled.

Fixes #12 

/cc @ialidzhikov 

**Release note**:

```feature user
The `ValidatingWebhookConfiguration` used to validate whether the `shoot-rsyslog-relp` extension is properly configured on `Shoot`s now has an `objectSelector` that will only select objects with the following label - `extensions.extensions.gardener.cloud/shoot-rsyslog-relp: 'true'`.
```

### DIFF
--- a/charts/gardener-extension-shoot-rsyslog-relp-admission/charts/application/templates/validatingwebhook-validator.yaml
+++ b/charts/gardener-extension-shoot-rsyslog-relp-admission/charts/application/templates/validatingwebhook-validator.yaml
@@ -20,7 +20,11 @@ webhooks:
     resources:
     - shoots
   failurePolicy: Fail
-  objectSelector: {}
+  objectSelector:
+    {{- if .Values.global.webhookConfig.useObjectSelector }}
+    matchLabels:
+      extensions.extensions.gardener.cloud/shoot-rsyslog-relp: "true"
+    {{- end }}
   namespaceSelector: {}
   sideEffects: None
   admissionReviewVersions:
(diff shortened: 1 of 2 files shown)


ialidzhikov (approved):
Thanks!

/lgtm

PR #17, opened 2023-10-02
Vendor `gardener@1.75.0` and drop dependency injection functions

AleksandarSavchev:
The package `controller-runtime/pkg/runtime/inject` has been removed in `controller-runtime` `v0.15`. This package contained long deprecated injection functions (like `InjectScheme`, `InjectLogger`, `InjectConfig`, `InjectClient`, `InjectCache`, etc.).

The runtime injection functionality has been deprecated since `Controller Runtime 0.10`; all of the above fields can be passed from the `Manager` to structs or interfaces that need them.

Fixes #

**Release note**:

```breaking operator
The `security.gardener.cloud/pod-security-enforce` annotation in the ControllerRegistration is set to `baseline`. With this, the pods running in the extension namespace should comply with `baseline` pod-security standard.
```


pkg/admission/validator/shoot.go
```diff
@@ -31,24 +32,15 @@ type shoot struct {
...
-func (s *shoot) InjectScheme(scheme *runtime.Scheme) error {
-	s.decoder = serializer.NewCodecFactory(scheme, serializer.EnableStrict).UniversalDecoder()
-	return nil
-}
-
-// InjectClient injects the given client into the validator.
-func (s *shoot) InjectClient(client client.Client) error {
-	s.client = client
-	return nil
+func NewShootValidator(mgr manager.Manager) extensionswebhook.Validator {
```
ialidzhikov, line 35: I would rather prefer to accept only the needed things instead of accepting a general `manager.Manager`. See https://github.com/gardener/gardener-extension-registry-cache/blob/5fdb7552139cc3a4ee98e460270b658ee0e188ab/pkg/admission/validator/shoot.go#L38-L42.

```suggestion
func NewShootValidator(client client.Client, decoder runtime.Decoder) extensionswebhook.Validator {
```

pkg/controller/lifecycle/actuator.go
```diff
@@ -45,8 +44,10 @@ const (
 )
 
 // NewActuator returns an actuator responsible for Extension resources.
-func NewActuator(config apisconfig.Configuration, chartRendererFactory extensionscontroller.ChartRendererFactory) extension.Actuator {
+func NewActuator(mgr manager.Manager, config apisconfig.Configuration, chartRendererFactory extensionscontroller.ChartRendererFactory) extension.Actuator {
```
ialidzhikov, line 47: Same as above.

```suggestion
func NewActuator(client client.Client, decoder runtime.Decoder, config apisconfig.Configuration, chartRendererFactory extensionscontroller.ChartRendererFactory) extension.Actuator {
```

Ref https://github.com/gardener/gardener-extension-registry-cache/blob/f3a362e9eb23c06cff2e2d6e800a395b3433c759/pkg/controller/extension/actuator.go#L43-L50

ialidzhikov (commented):
Thanks for the PR! Minor nits inline.

ialidzhikov (approved):
/lgtm

plkokanov (approved):

PR #19, opened 2023-10-11
Adds PriorityClass to daemonsets' pods deployed in the shoot cluster

### LINKED ISSUE #16: Set priority class to shoot system rsyslog pods

`rsyslog-relp-configurator` and `rsyslog-relp-configuration-cleaner` pods should have priorityclass set similar to other gardener system components so that they can be scheduled with priority compared to less important pods on the node.

plkokanov:
This PR adds a priority class to the `rsyslog-configurator` and `rsyslog-configuration-cleaner` pods so that they can be properly scheduled. This was previously missed to be added.

 I have chosen `gardener-shoot-system-700` so that the rsyslog-relp pods do not preempt the `node-problem-detector` and other critical `daemonsets`. Also it makes sense to have the same priority as other log/metric exporters.

Fixes #16 

**Release note**:

```bugfix operator
Added the `gardener-shoot-system-700` priority class to the `rsyslog-relp-configurator` and `rsyslog-relp-configuration-cleaner` pods.
```

### DIFF
--- a/charts/internal/rsyslog-relp-configuration-cleaner/templates/daemonset.yaml
+++ b/charts/internal/rsyslog-relp-configuration-cleaner/templates/daemonset.yaml
@@ -19,6 +19,7 @@ spec:
       labels:
 {{ include "labels" . | indent 8 }}
     spec:
+      priorityClassName: gardener-shoot-system-700
       containers:
       - name: pause-container
         image: {{ .Values.images.pause }}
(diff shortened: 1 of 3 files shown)


ialidzhikov (approved):
/lgtm

ialidzhikov:
You could also update the docs in gardener/gardener similar to https://github.com/gardener/gardener/pull/8554.

Issue #2, opened 2023-08-08
☂️ Getting gardener-extension-shoot-rsyslog-relp production ready


plkokanov:
**What would you like to be added**:
/area security audit-logging
/kind task

**What would you like to be added**:
Tasks to make the gardener-extension-shoot-rsyslog-relp production ready.

**Why is this needed**:
Getting the shoot-rsyslog-relp extension production ready.

## Steps

- [x] Add prow jobs to execute checks and tests, validate PRs and manage releases: https://github.com/gardener/ci-infra/pull/857. @plkokanov
- [x] Use `skaffold` to deploy the extension when developing locally instead of the [rsyslog-relp-up-kind.sh](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/d59c9bccf67d4ed9ae827338748e8ea57bf8ec69/Makefile#L137-L139) script: https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/pull/5
- [x] https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/issues/12
- [x] https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/issues/16 https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/pull/19
- [x] Add monitoring config for rsyslog (and auditd) to prometheus in shoot control plane 
  - [x] The `node-exporter` can be reused if its textfile collector is enabled. The metrics from rsyslog can be transformed by a script so that they are in a format that `node-exporter` accepts and then written to a `*.prom` file which can be read by the textfile collector.
    - [x] Prepare a POC and sync it with Istvan https://github.com/plkokanov/gardener-extension-shoot-rsyslog-relp/tree/poc/prometheus-metrics
    - [x] Enable the `textfile-collector` of the `node-exporter` in g/g https://github.com/gardener/gardener/pull/8721
    - [x] Expose rsyslog metrics via node-exporter's textfile collector and display them in a dedicated plutono dashboard https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/pull/32
  -  Two other options currently have been evaluated, but for now the textfile seems better: 
        1.Possibly explore how metrics are exposed via https://github.com/prometheus-community/rsyslog_exporter
        2. A different option would be to reuse a scraper running on the node which simply reads from the `/var/log/rsyslog-stats.log` to which rsyslog is configured to write metrics.
[...]

PR #27, opened 2023-10-22
Makes it possible to configure the `tls.tlslib` option for the relp output module.

plkokanov:
This PR makes it possible to configure the `tls.tlslib` option for the relp output module via the configuration of the `shoot-rsyslog-extension`. This option determines which library is used by `librelp` when the connection to the target server is encrypted via TLS. The two options currently are `openssl` and `gnutls`. Note that not all operating systems support both libraries. If `tls.tlslib` is not specified, the default tls library will be used by `librelp`, in most cases this is `gnutls`. More information can be found here: https://www.rsyslog.com/doc/v8-stable/configuration/modules/imrelp.html#tls-tlslib

In most cases using `gnutls` should be ok. However, it seems that when using certificates generated on MacOS (even with gnu `certtool`)  and `librelp` uses `gnutls`, the RELP connection fails with the following errors:
```
Oct 22 13:32:07 machine-shoot--local--local-local-7d9f7-lcpqb rsyslogd[2167334]: omrelp[10.2.64.54:443]: error 'TLS record reception failed [gnutls error -110: The TLS connection was non-properly terminated.]', object  'conn to srvr 10.2.64.54:443' - action may not work as intended [v8.2102.0 try https://www.rsyslog.com/e/2353 ]
```
Switching to `openssl` fixes this problem.

This PR also adds an E2E test which tests the extension when the RELP connection uses tls. Specifying the `openssl` library to be used by `librelp` was necessary to make the tests work. 

Part of #2 

**Release note**:

```feature user
The `shoot-rsyslog-relp` configuration now allows users to specify which tls library should be used by `librerlp` when tls communication is enabled via the `tls.tlsLib` optional field. The possible options are `gnutls` and `openssl`. When the field is omitted, `librelp` uses its default tls library which in most cases is `gnutls`. More information can be found here: https://www.rsyslog.com/doc/v8-stable/configuration/modules/imrelp.html#tls-tlslib
```


example/local/01-shoot.yaml
```diff
@@ -34,8 +35,15 @@ spec:
     #   enabled: true
     #   secretReferenceName: rsyslog-tls-certificates
     #   authMode: name # {name, fingerprint}
+    #   tlsLib: openssl # {openssl, gnutls}
```
ialidzhikov, line 38: If `openssl` works everywhere, then we could consider using it internally instead of exposing it as configuration option. I don't know in which cases you (an end user) would need to specify `gnutls`? The docs (https://www.rsyslog.com/doc/v8-stable/configuration/modules/imrelp.html#tls-tlslib) recommend using `openssl`.

docs/usage/configuration.md
```diff
@@ -54,16 +54,19 @@ spec:
...
-      # Add tls settings if tls should be used to encrypt th connection to the target server.
+      # Add tls settings if tls should be used to encrypt the connection to the target server.
       tls:
         enabled: true
         # Use `name` authentication mode for the tls connection.
         authMode: name
         # Only allow connections if the server's name is `some.rsyslog-rlep.server`
-        permittedPeer: some.rsyslog-rlep.server
+        permittedPeer:
+        - "some.rsyslog-rlep.server"
```
ialidzhikov, line 64: Nit: there is also occurrence in the README.md that can be fixed:
```diff
index 08712e3..1729791 100644
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ port: 1520
 tls:
   enabled: true
   authMode: name
-  permittedPeer: rsyslog-server-peer
+  permittedPeer:
+  - rsyslog-server-peer
   secretReferenceName: rsyslog-relep-tls
```

test/e2e/create_enable_disable_delete.go
```diff
@@ -12,68 +12,186 @@ import (
...
-	f.Shoot = e2e.DefaultShoot("e2e-rsyslog-relp")
+	test := func(shoot *gardencorev1beta1.Shoot, shootMutateFn func(shoot *gardencorev1beta1.Shoot) error, gardenClusterResources ...client.Object) {
+		f := defaultShootCreationFramework()
+		f.Shoot = shoot
+
+		It("Create Shoot, enable shoot-rsyslog-relp extension then disable it and delete Shoot", Offset(1), func() {
+			ctx, cancel := context.WithTimeout(parentCtx, 20*time.Minute)
+			DeferCleanup(cancel)
+
+			By("Crerate additional garden cluster resources")
```
ialidzhikov, line 30:
```suggestion
			By("Create additional garden cluster resources")
```

test/e2e/create_enable_disable_delete.go
```diff
@@ -12,68 +12,186 @@ import (
...
+					},
+				})
+				return nil
+			},
+			&corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "rsyslog-relp-tls",
+				},
+				Data: map[string][]byte{
+					"ca": []byte(`-----BEGIN CERTIFICATE-----
```
ialidzhikov, line 123: Idea: You could move it to files under `$PWD/testdata` and load them with `go:embed` for the test? WDYT?

test/e2e/create_enable_disable_delete.go
```diff
@@ -12,68 +12,186 @@ import (
...
-			return nil
-		})).To(Succeed())
+		})
+	}
 
-		By("Verify that shoot-rsyslog-relp extension is disabled")
-		ctx, cancel = context.WithTimeout(parentCtx, 5*time.Minute)
-		defer cancel()
-		verifier.verifyThatLogsAreNotSentToEchoServer(ctx, "test-program", "1", "this should not get sent to echo server")
+	Context("shoot-rsyslog-relp extension with tls disabled", Label("tls-disabled"), func() {
```
ialidzhikov, line 96: Idea: instead of passing a Shoot and mutate func and gardenClusterResources to be created, you could make the `test` func a helper func (without the IT inside). The test func can accept the final Shoot and you can set up the needed Shoot and created the needed resources a BeforeEach or as first step in ITs. WDYT?

example/local/01-shoot.yaml
plkokanov: The problem is that `openssl` doesn't work everywhere :(. It depends on what library was used to generate the certificates, what tls library is currently available on the system, the version of `librelp` and probably other factors.

test/e2e/create_enable_disable_delete.go
plkokanov: I wanted to test the enablement and disablement of the extension in an already created shoot, that's why I use the mutate func.

If the `test()` func does not contain an `It()` but the other way around, I would have to code it like this:

```go
Context("shoot-rsyslog-relp extension with tls disabled", func() {
  	f := defaultShootCreationFramework()
	f.Shoot = e2e.DefaultShoot("e2e-rsyslog-relp")
	
	It("Create Shoot, enable shoot-rsyslog-relp extension then disable it and delete Shoot", func() {
		test(f, ...)
        })   
})

Context("shoot-rsyslog-relp extension with tls disabled", func() {
  	f := defaultShootCreationFramework()
	f.Shoot = e2e.DefaultShoot("e2e-rsyslog-relp")
	BeforeEach(func() {
                //create resources
        })

	It("Create Shoot, enable shoot-rsyslog-relp extension then disable it and delete Shoot", func() {
                 // optionally create resources here
		test(f, ...)
        })   
})
```

If we want to run the 2 cases in Parallel, we need to use 2 separate `ShootCreationFrameWork`s. Additionally, the  `defaultShootCreationFramework()` function will call a few `BeforeEach()` functions so it cannot be used within an `It()`.

Initially I had it coded like this:
```go
test = func(f *ShootCreationFramework) {
	It("Create Shoot, enable shoot-rsyslog-relp extension then disable it and delete Shoot", func() {...})
}

Context("shoot-rsyslog-relp extension with tls disabled", func() {
  	f := defaultShootCreationFramework()
	f.Shoot = e2e.DefaultShoot("e2e-rsyslog-relp")
	
	test(f, mutateFunc)
})

Context("shoot-rsyslog-relp extension with tls disabled", func() {
  	f := defaultShootCreationFramework()
	f.Shoot = e2e.DefaultShoot("e2e-rsyslog-relp")
	BeforeEach(func() {
                // We need access to `f` here to create resources
        })

	test(f, mutateFunc)  
})
```

But I wanted to avoid the duplicated code for the creation of the `ShootCreationFramework`

test/e2e/create_enable_disable_delete.go
plkokanov: Thanks for the suggestion, I found a function in `"github.com/gardener/gardener/pkg/utils/test.EnsureTestResources` so I can add the whole resource in a `.yaml` file and use that function to deploy it

ialidzhikov (approved):
/lgtm

PR #28, opened 2023-10-25
Removes the `healthcheck` controller

plkokanov:

/kind cleanup

After https://github.com/gardener/gardener/pull/7462 gardenlet performs a health check for all ManagedResources (without the ones with class seed) in the Shoot control plane. The shoot-rsyslog-relp extension deploys only a single ManagedResource when the extension is enabled, and a single ManagedResource (which performs clean-up activities) when the extension is disabled. It needs to report health checks for them. We no longer need to have a health check controller to do this. We can rely that gardenlet will health check the ManagedResource deployed by the extension.

Part of #2 

/cc @ialidzhikov as this is similar to https://github.com/gardener/gardener-extension-registry-cache/pull/32

**Release note**:

```other operator
The healthcheck controller is now removed. Starting [v1.65.0](https://github.com/gardener/gardener/releases/tag/v1.65.0), gardenlet perform health checks for all ManagedResources in the Shoot control plane in the Seed. There is no longer need of the custom healthcheck controller in the shoot-rsyslog-relp extension as it was doing the same job. It was performing health check for the ManagedResource it deploys.
```


pkg/controller/lifecycle/actuator.go
```diff
@@ -166,6 +167,7 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, ex *extensionsv
 			"alpine": imagevector.AlpineImage(),
 			"pause":  imagevector.PauseContainerImage(),
 		},
+		"pspDisabled": gardencorev1beta1helper.IsPSPDisabled(cluster.Shoot),
```
ialidzhikov, line 170: Is it used currently? I assume you want in a dedicate PR add the needed PSPs?

ialidzhikov (commented):
Looks good! Minor nit only.

ialidzhikov (approved):
/lgtm

PR #30, opened 2023-11-01
Do not deploy the rsyslog-relp-configuration-cleaner on Shoot deletion

plkokanov:

/kind enhancement

Similrly to https://github.com/gardener/gardener-extension-registry-cache/pull/83, if the Shoot is in deletion, then there is no need to clean up the rsyslog configuration from Nodes. The Shoot deletion flow ensures that the Worker is deleted before the Extension deletion.
https://github.com/gardener/gardener/blob/6f8ce7692f793e583734f533cfdee7b9d93fb815/pkg/gardenlet/controller/shoot/shoot/reconciler_delete.go#L436-L440

The `deleteExtensionResourcesBeforeKubeAPIServer` step depends on `waitUntilManagedResourcesDeleted` which depends on `deleteManagedResources` which depends on `waitUntilWorkerDeleted`.

Fixes #

**Release note**:

```other operator
The `rsyslog-relp-configuration-cleaner` is no longer deployed on Shoot deletion with `shoot-rsyslog-relp` extension enabled. The Extension deletion occurs after the Worker deletion. There are no Nodes, hence there is no need to clean up registry configuration.
```

### DIFF
--- a/pkg/controller/lifecycle/actuator.go
+++ b/pkg/controller/lifecycle/actuator.go
@@ -200,6 +200,13 @@ func (a *actuator) Delete(ctx context.Context, _ logr.Logger, ex *extensionsv1al
 		return err
 	}
 
+	// If the Shoot is in deletion, then there is no need to clean up the rsyslog configuration from Nodes.
+	// The Shoot deletion flows ensures that the Worker is deleted before the Extension deletion.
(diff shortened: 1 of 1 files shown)


acumino (commented):
/lgtm

ialidzhikov (approved):
/lgtm

PR #33, opened 2023-11-10
Update `pause` and `alpine` images and skip scans for the `alpine` image

plkokanov:

/kind enhancement

This PR updates the  `pause` and `alpine`  container images to `3.9` and `3.18.4`,  respectively.
It also labels the `alpine` image to be skipped from security scans as it is not accessible from outside k8s clusters and not interacted with from other containers or other systems. This is the same as the reason to skip the scans of the `pause` container.

Fixes #

**Release note**:

```other operator
Vulnerability scans are disabled for the alpine image as the corresponding container is not accessible from outside of the k8s clusters and not interacted with from other containers or other systems.
```
```other operator
The following images are updated:
- `eu.gcr.io/gardener-project/3rd/alpine`: 3.15.8 -> 3.18.4
- `registry.k8s.io/pause`: 3.7 -> 3.9
```

### DIFF
--- a/charts/images.yaml
+++ b/charts/images.yaml
@@ -6,16 +6,23 @@
 images:
 - name: alpine
   repository: eu.gcr.io/gardener-project/3rd/alpine
-  tag: "3.15.8"
-
-- name: pause-container
-  sourceRepository: github.com/kubernetes/kubernetes/blob/master/build/pause/Dockerfile
-  repository: registry.k8s.io/pause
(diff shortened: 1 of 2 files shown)


ialidzhikov (approved):
/lgtm

ialidzhikov:
Same as https://github.com/gardener/gardener-extension-registry-cache/pull/89

PR #37, opened 2023-11-20
Revendor `gardener/gardener` v1.84.0

plkokanov:

/kind enhancement

Revendor `gardener/gardener` `v1.84.0`

Fixes #

**Release note**:

```other operator

```


go.mod
```diff
@@ -5,10 +5,10 @@ go 1.21
 require (
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/ahmetb/gen-crd-api-reference-docs v0.3.0
-	github.com/gardener/gardener v1.81.6
+	github.com/gardener/gardener v1.84.0
```
ialidzhikov, line 8: You could also do the following change - ref https://github.com/gardener/gardener-extension-registry-cache/pull/97/commits/09e61ff3878f76b0af1518b142a51b8f4be9b350.

plkokanov: Ah, nice thanks for reminding me, completely forgot about that change :)

plkokanov: There's an issue now that the `vendor` folder has been removed from gardener in `v1.84.0` where the `github.com/gardener/gardener/hack/check-skaffold-deps-for-binary.sh` script would also not add the `vendor` folder to the skaffold dependencies leading to the following error:
```
>>> The following dependencies are not needed actually (need to be removed):
vendor
```

We could merge the revendoring PR now and only adapt to using the `github.com/gardener/gardener/hack/check-skaffold-deps-for-binary.sh` script after the `vendor` directory is also removed from this extension.
This also has the caveat that `check-generate` will no longer detect if `make revendor` has not been called, but I think we can live with that for now.

ialidzhikov: I see. I now added a new subtask to #2:
> - [ ] Use the `hack/check-skaffold-deps-for-binary.sh` from Gardener scripts. Similar to https://github.com/gardener/gardener-extension-registry-cache/pull/97/commits/09e61ff3878f76b0af1518b142a51b8f4be9b350

ialidzhikov (approved):
/lgtm

plkokanov:
I will use this PR to revendor gardener `v1.84.1` and make the necessary changes to the skaffold checks.

plkokanov:
@ialidzhikov I modified the PR and vendored `gardener@1.84.1`. I also update the `check-skaffold-deps.sh` script. PTAL

ialidzhikov (approved):
Thanks!

/lgtm

PR #29, opened 2023-10-27
Align `shoot-rsyslog-relp` extension with `gardener`'s component checklist

plkokanov:
**How to categorize this PR?**

/kind task 
/kind cleanup

**What this PR does / why we need it**:
This PR covers part of the items from [component-checklist](https://github.com/gardener/gardener/blob/master/docs/development/component-checklist.md) listed in #2
More concretely it:
- Reduces RBAC for the extension controller
- Adds proper `high-availability-config.resources.gardener.cloud/type` label for the extension controller
- Removes the `securityContext` from the admission deployment, as that is now handled by a webhook in GRM
- Adds well formed `securityContext` to the daemonsets deployed in the shoot
- Adds `PodSecurityPolicies` for the daemonsets deployed in the shoot in case the shoot cluster is still using k8s version `1.24` and does not disable the PSP admission plugin
- Adds resource requests and limits to the init containers of the daemonsets deployed in the shoot. The resources requests/limits are only added on the init containers as it does not make much sense to add  them to the containers runnning the pause image
- Adapts integration tests to the new changes - mainly the addition of PSPs
- Not part of the component checklist, but worth mentioning here - the PR removes some no-longer needed make targets from the project's `Makefile`.

```
General
❌ 1. Avoid usage of Helm charts: This will be done in a dedicated PR.
✅ 2. Choose the proper deployment way: We use ManagedResource to deploy the `rsyslog-relp-configurator` and the `rsyslog-relp-configuration-cleaner` to the Shoot.
✅ 3. Do not hard-code container image references: We don't - ref https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/main/charts/images.yaml
✅ 4. Do not use container images from Docker Hub: We don't - ref https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/main/charts/images.yaml
❌ 5. Use unique ConfigMaps/Secrets: This will be done in a separate PR when we remove helm charts
✅ 6. Manage certificates/secrets via secrets manager: We don't have such Secret.
✅ 7. Consider hibernation when calculating replica count: We don't have a workload that is deployed to the Shoot control plane.
✅ 8. Ensure task dependencies are as precise as possible in shoot flows: Not applicable.
✅ 9. Handle shoot system components: Looks fine for me.

Security
✅ 1. Use a dedicated ServiceAccount and disable auto-mount: The pods deployed in the Shoot don't talk to the kube-apiserver.
[...]

PR #39, opened 2023-12-05
Restructure examples, update README and add missing licenses

plkokanov:

/kind enhancement

This PR restructures the `example` dir and removes some no longer necessary files in line with the changes to the `Makefile` from https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/pull/29
Additionally, the PR adds some missing license headers and adapts the `README` so that it is similar to the one https://github.com/gardener/gardener-extension-registry-cache 
 
Part of #2 

**Release note**:

```other operator
The name of the `gardener-extension-shoot-rsyslog-relp-admission` chart is now correctly specified as `gardener-extension-shoot-rsysloog-relp-admission`. Previously it was `gardener-extension-shoot-rsyslog-relp`. This should not require anything to be done by operators when upgrading the chart.
```


charts/gardener-extension-shoot-rsyslog-relp-admission/Chart.yaml
```diff
@@ -1,5 +1,10 @@
+# SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+---
 apiVersion: v1
 appVersion: "1.0"
 description: A Helm chart for the Gardener Shoot Rsyslog Relp admission controller
-name: gardener-extension-shoot-rsyslog-relp
+name: gardener-extension-shoot-rsyslog-relp-admission
```
ialidzhikov, line 9: I was wondering whether this is a breaking change? If someone already had the chart deployed with help, does he/she need something to do to switch to the new name?

test/integration/controller/lifecycle/resources/crd-resources.gardener.cloud_managedresources.yaml
```diff
@@ -240,4 +240,4 @@ spec:
     served: true
     storage: true
     subresources:
-      status: {}
+      status: {}
```
ialidzhikov, line 243: Some other extensions are generating these CRDs based on the vendored version of gardener/gardener. See https://github.com/gardener/gardener-extension-provider-aws/blob/f9edf4533806c8885bdeaa178aba6dea09b98eaf/example/doc.go#L15

You could do the same instead of maintaining them manually.

ialidzhikov (changes requested):

example/local/admission/networkpolicy.yaml
```diff
@@ -1,3 +1,8 @@
+# SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
 apiVersion: networking.k8s.io/v1
```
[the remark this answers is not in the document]
plkokanov: The labels already exist https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/fae912cf517a2abb86095ffb44cb13825d0c0c4c/charts/gardener-extension-shoot-rsyslog-relp-admission/charts/runtime/templates/deployment.yaml#L27-L29

For some reason the local setup does not work without this network policy. 

I know that it works without such a policy for the registry-cache extension, however I did not have much time to see what the exact reason for the different behaviour is.

charts/gardener-extension-shoot-rsyslog-relp-admission/Chart.yaml
plkokanov: Good point, I think I tried upgrading it in the local setup with skaffold - first ran `make extensions-up` with the old release name, then `make extensions-up` with the new name, however let me try it out again and perform some more checks.

example/local/admission/networkpolicy.yaml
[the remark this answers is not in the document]
ialidzhikov: It works without the NetworkPolicy according to my testing: ref https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/pull/46

charts/gardener-extension-shoot-rsyslog-relp-admission/Chart.yaml
plkokanov: The only difference should be the chart's name when the `runtime` and `application` subcharts are installed in the same cluster. In that case one would usually install the parent chart directly. Helm handles the rename of the charts without a problem and since `.Chart.Name` is not used anywhere and there are no chart dependencies there should be no additional work that has to be done.

plkokanov: I anyway added a release note about this change.

test/integration/controller/lifecycle/resources/crd-resources.gardener.cloud_managedresources.yaml
ialidzhikov: I added a new subtask to https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/issues/2. When the provider extensions tackle the above described issue, then we can also follow their path.

PR #41, opened 2023-12-07
Deploy `rsyslog` configuration files via a webhook that mutates `OperatingSystemConfig` resources

plkokanov:

/kind enhancement

This PR changes the way that the `rsyslog` configuration files are deployed so that an `OperatingSystemConfig` mutating webhook is used. The files are placed in the same locations as with the previous deployment via the `rsyslog-relp-configurator` daemonset.

Additionally, if any shoots were created with the old daemonset approach, the corresponding `ManagedResource` is deleted as part of the shoot reconciliation. This logic can be dropped later on

Part of #2 

**Release note**:

```other operator
The extension now deploys the `rsyslog` configuration files by mutating the `OperatingSystemConfig` resource via a mutating webhook. Cleanup of the `rsyslog` configuration files is still handled by the `rsyslog-relp-configuration-cleaner` daemonset.
```


pkg/webhook/operatingsystemconfig/auditd.go
```diff
@@ -0,0 +1,87 @@
...
+	systemIntegrityRules = `-a exit,always -F dir=/boot -F perm=wa -F key=system_integrity
+-a exit,always -F dir=/etc -F perm=wa -F key=system_integrity
+-a exit,always -F dir=/bin -F perm=wa -F key=system_integrity
+-a exit,always -F dir=/sbin -F perm=wa -F key=system_integrity
+-a exit,always -F dir=/lib -F perm=wa -F key=system_integrity
+-a exit,always -F dir=/lib64 -F perm=wa -F key=system_integrity
+-a exit,always -F dir=/usr -F perm=wa -F key=system_integrity
+-a exit,always -F dir=/opt -F perm=wa -F key=system_integrity
+-a exit,always -F dir=/root -F perm=wa -F key=system_integrity`
+)
```
ialidzhikov, line 42: As these are actually files on the host, it might be more intuitive to have them as files and the read them with go:embed. WDYT?

pkg/controller/lifecycle/actuator.go
```diff
@@ -74,113 +67,14 @@ func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, ex *extensionsv
...
-		"images": map[string]interface{}{
-			"alpine": imagevector.AlpineImage(),
-			"pause":  imagevector.PauseContainerImage(),
-		},
-		"pspDisabled": gardencorev1beta1helper.IsPSPDisabled(cluster.Shoot),
-	}
-
-	release, err := chartRenderer.RenderEmbeddedFS(charts.InternalChart, charts.RsyslogConfiguratorChartPath, releaseName, metav1.NamespaceSystem, values)
-	if err != nil {
+	// TODO(plkoknanov): remove this after a couple releases.
```
ialidzhikov, line 70:
```suggestion
	// TODO(plkokanov): remove this after a couple releases.
```

ialidzhikov, line 70: L61-L68 seems no longer needed, you can drop it:
https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/b0b214bd593fd46f4fe571b68cc04e9a4edb7d9c/pkg/controller/lifecycle/actuator.go#L61-L68

ialidzhikov (commented):
Overall lgtm, minor suggestions inline.

pkg/webhook/operatingsystemconfig/auditd.go
plkokanov: I initially had them as files, but decided to bring them in as constants because I couldn't come up with a good folder structure. However, I think something like this could do the job:
```
pkg/webhook/operatingsystemconfig/resources/
├── auditrules
│   ├── 00-base-config.rules
│   └── 10-privilege-escalation.rules
└── templates
    ├── 60-audit.conf.tpl
    └── scripts
        ├── configure-rsyslog.tpl.sh
        └── process-rsyslog-pstats.tpl.sh
```

ialidzhikov: I like it. Dedicated folder for audit rules makes sense!

pkg/webhook/operatingsystemconfig/rsyslog.go
```diff
@@ -0,0 +1,320 @@
...
+			Namespace: namespace,
+		},
+	}
+	if err = c.Get(ctx, client.ObjectKeyFromObject(refSecret), refSecret); err != nil {
+		return nil, err
+	}
+
+	if err := utils.ValidateRsyslogRelpSecret(refSecret); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
plkokanov: As discussed, lets keep the validation for now, since there is no validation if the referenced secret is updated in the `shoot-rsyslog-relp-admission`. Instead I've added a `TODO` to remove the validation once the referenced secret is forced to be immutable and also a subtask for this in #2

ialidzhikov (approved):
/lgtm

dimitar-kostadinov (commented):
/lgtm

PR #42, opened 2023-12-08
Aligns image target names with image registry name

plkokanov:
**How to categorize this PR?**

/kind bug

**What this PR does / why we need it**:
This PR aligns the target names in the repository's `Dockerfile` with the names of the images' registry. 
This is now necessary because there is no option to explicitly specify the image repository in prow (unlike what was configured in concourse [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/2f627b997dfe73aec65903ad0d0415dd904c1b16/.ci/pipeline_definitions#L48)) and therefore [this configuration](https://github.com/gardener/ci-infra/blob/master/config/jobs/gardener-extension-shoot-rsyslog-relp/gardener-extension-shoot-rsyslog-relp-build-images.yaml) will cause the images built from head updates to be available at `eu.gcr.io/gardener-project/gardener/extensions/gardener-extension-{shoot-rsyslog-relp,shoot-rsyslog-relp-admission}` instead of `eu.gcr.io/gardener-project/gardener/extensions/{shoot-rsyslog-relp,shoot-rsyslog-relp-admission}` where they were created by concourse.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```other operator
NONE
```

PR #45, opened 2023-12-11
Fix `make docker-images` target and imporve the Makefile

plkokanov:

/kind bug

The PR does the following fixes and improvements:
- Fixes the image targets used in the `make docker-images` command after they were changed in #42.
- Additionally, it improves the Makefile by removing some unnecessary variables and properly passing the `LD_FLAGS` variable to `skaffold`
- `GOARCH=$TARGETARCH` was removed from the `Dockerfille` as it is anyway handled by the `$(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/install.sh` script: https://github.com/gardener/gardener/blob/b83a3a0932698ebf267462dff30f3e31c8f912ac/hack/install.sh#L23

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+EXTENSION_PREFIX            := gardener-extension
 NAME                        := shoot-rsyslog-relp
 NAME_ADMISSION              := $(NAME)-admission
 IMAGE                       := eu.gcr.io/gardener-project/gardener/extensions/shoot-rsyslog-relp
@@ -10,7 +11,7 @@ HACK_DIR                    := $(REPO_ROOT)/hack
 VERSION                     := $(shell cat "$(REPO_ROOT)/VERSION")
 EFFECTIVE_VERSION           := $(VERSION)-$(shell git rev-parse HEAD)
 IMAGE_TAG                   := $(EFFECTIVE_VERSION)
-LD_FLAGS                    := "-w $(shell EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) $(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/get-build-ld-flags.sh k8s.io/component-base $(REPO_ROOT)/VERSION $(NAME))"
+LD_FLAGS                    := "-w $(shell EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) $(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/get-build-ld-flags.sh k8s.io/component-base $(REPO_ROOT)/VERSION $(EXTENSION_PREFIX)-$(NAME))"
 PARALLEL_E2E_TESTS          := 2
 
 ifndef ARTIFACTS
--- a/Makefile
+++ b/Makefile
@@ -2,16 +2,15 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-EXTENSION_PREFIX            := gardener-extension
-NAME_ADMISSION              := shoot-rsyslog-relp-admission
 NAME                        := shoot-rsyslog-relp
-REGISTRY                    := eu.gcr.io/gardener-project/gardener
-IMAGE_PREFIX                := $(REGISTRY)/extensions
+NAME_ADMISSION              := $(NAME)-admission
+IMAGE                       := eu.gcr.io/gardener-project/gardener/extensions/shoot-rsyslog-relp
 REPO_ROOT                   := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 HACK_DIR                    := $(REPO_ROOT)/hack
 VERSION                     := $(shell cat "$(REPO_ROOT)/VERSION")
 EFFECTIVE_VERSION           := $(VERSION)-$(shell git rev-parse HEAD)
-GOARCH                      ?= $(shell go env GOARCH)
+IMAGE_TAG                   := $(EFFECTIVE_VERSION)
+LD_FLAGS                    := "-w $(shell EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) $(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/get-build-ld-flags.sh k8s.io/component-base $(REPO_ROOT)/VERSION $(NAME))"
 PARALLEL_E2E_TESTS          := 2
 
 ifndef ARTIFACTS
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -35,17 +34,17 @@ include $(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/tools.mk
 
 .PHONY: install
 install:
-	@LD_FLAGS="-w $(shell EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) $(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/get-build-ld-flags.sh k8s.io/component-base $(REPO_ROOT)/VERSION $(EXTENSION_PREFIX)-$(NAME))" \
-	$(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/install.sh ./...
+	@LD_FLAGS=$(LD_FLAGS) \
+	$(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/install.sh ./cmd/...
```
shafeeqes, line 38: Why do we need this change?

Makefile
```diff
@@ -2,16 +2,15 @@
...
-IMAGE_PREFIX                := $(REGISTRY)/extensions
+NAME_ADMISSION              := $(NAME)-admission
+IMAGE                       := eu.gcr.io/gardener-project/gardener/extensions/shoot-rsyslog-relp
 REPO_ROOT                   := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 HACK_DIR                    := $(REPO_ROOT)/hack
 VERSION                     := $(shell cat "$(REPO_ROOT)/VERSION")
 EFFECTIVE_VERSION           := $(VERSION)-$(shell git rev-parse HEAD)
-GOARCH                      ?= $(shell go env GOARCH)
+IMAGE_TAG                   := $(EFFECTIVE_VERSION)
+LD_FLAGS                    := "-w $(shell EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) $(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/get-build-ld-flags.sh k8s.io/component-base $(REPO_ROOT)/VERSION $(NAME))"
```
shafeeqes, line 13: Instead of NAME, Can we not use `$EXTENSION_PREFIX-$NAME`? I see that in other extensions we only use prefix, but name also could be added.

Makefile
```diff
@@ -35,17 +34,17 @@ include $(REPO_ROOT)/vendor/github.com/gardener/gardener/hack/tools.mk
...
 .PHONY: docker-login
 docker-login:
 	@gcloud auth activate-service-account --key-file .kube-secrets/gcr/gcr-readwrite.json
 
 .PHONY: docker-images
 docker-images:
-	@docker build --build-arg EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) --build-arg TARGETARCH=$(GOARCH) -t $(IMAGE_PREFIX)/$(NAME):$(EFFECTIVE_VERSION) -t $(IMAGE_PREFIX)/$(NAME):latest -f Dockerfile -m 6g --target $(EXTENSION_PREFIX)-$(NAME) .
-	@docker build --build-arg EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) --build-arg TARGETARCH=$(GOARCH) -t $(IMAGE_PREFIX)/$(NAME_ADMISSION):$(EFFECTIVE_VERSION) -t $(IMAGE_PREFIX)/$(NAME_ADMISSION):latest -f Dockerfile -m 6g --target $(EXTENSION_PREFIX)-$(NAME_ADMISSION) .
+	@docker build --build-arg EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) -t $(IMAGE):$(IMAGE_TAG) -f Dockerfile -m 6g --target $(NAME) .
+	@docker build --build-arg EFFECTIVE_VERSION=$(EFFECTIVE_VERSION) -t $(IMAGE)-admission:$(IMAGE_TAG) -f Dockerfile -m 6g --target $(NAME_ADMISSION) .
```
shafeeqes, line 47: Don't we also need the `latest` tag?

plkokanov: I decided to skip it for these commands as they are mainly used for local builds. The `latest` tag will be added when building in prow: [ref](https://github.com/gardener/ci-infra/blob/e2802d3a85c87a064e1137c06907f7b64991b20a/config/jobs/gardener-extension-shoot-rsyslog-relp/gardener-extension-shoot-rsyslog-relp-build-images.yaml#L30) and I didn't want it to get overriden by mistake from local builds.
It is the same in [the registry-cache extension](https://github.com/gardener/gardener-extension-registry-cache/blob/a78ccd730f9dea3b43c32bfa6f21d231bfcc4f8a/Makefile#L53)
However, I could still add if you think that is the correct approach.

Makefile
plkokanov: I exported `LD_FLAGS` as a variable so that it can also be used [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/cd0ede12d77ff02d9c97e63bcd13c902da9e1b71/Makefile#L125)
In gardener we export it like [this](https://github.com/gardener/gardener/blob/b83a3a0932698ebf267462dff30f3e31c8f912ac/Makefile#L316)
It's not exported as a variable there, because I think we do not need it when running `make install` in g/g: [ref](https://github.com/gardener/gardener/blob/b83a3a0932698ebf267462dff30f3e31c8f912ac/Makefile#L89)

shafeeqes: I was rather asking about the `./cmd/...` change.

Makefile
shafeeqes: No strong opinions. We can keep it this way.

Makefile
plkokanov: I copied this also from the `registry-cache` extension: [ref](https://github.com/gardener/gardener-extension-registry-cache/blob/a78ccd730f9dea3b43c32bfa6f21d231bfcc4f8a/Makefile#L24)
TBH, I'm not 100% sure what should be the correct program name here, just using the extension prefix seems weird to me.

shafeeqes: I would prefer `gardener-extension-shoot-rsyslog-relp` to be printed when we call the `verflag.PrintAndExitIfRequested()` function rather than `shoot-rsyslog-relp`. But I leave the choice to you.

plkokanov: Yeah, the main problem is that it's the same for the admission, but we have this "issue" everywhere. Anyway, I will use your suggestion and set it to `gardener-extension-shoot-rsyslog-relp`

shafeeqes: > it's the same for the admission, but we have this "issue" everywhere

True.

plkokanov:
Thanks a lot for the review @shafeeqes! I addressed your comments.

shafeeqes:
Thanks!
/lgtm

PR #58, opened 2024-01-08
Ensure that `rsyslog.service` is enabled before configuring it

plkokanov:

/kind bug

This PR fixes a bug where `rsyslog.service` might not run properly if it was not enabled before. When `rsyslog.service` is not enabled, the `/etc/systemd/system/syslog.service` symlink is not created which creates an invalid system logging setup.
This issue was affecting operating systems on which `rsyslog.service` was not enabled by default, e.g. `gardenlinux`.

With this change, it should not be necessary to check if `/etc/systemd/system/syslog.service` exists before configuring `rsyslog.service`.

Fixes #

**Release note**:

```other operator
Fixed an issue where `rsyslog.service` would never get enabled if it was not already enabled by default.
```


dimitar-kostadinov (commented):
/lgtm

PR #60, opened 2024-01-10
Move `rsyslog-relp-echo-server` image to artifact registry

plkokanov:

/area TODO
/kind TODO

This PR adapts the testmachinery test so that the `rsyslog-relp-echo-server` image is fetched from the artifact registry: `europe-docker.pkg.dev/gardener-project/releases/gardener/extensions/shoot-rsyslog-relp-echo-server`

Additionally, it changes the repository for the image builds in the `Makefile` from `europe-docker.pkg.dev/gardener-project/public/gardener/extensions` to `europe-docker.pkg.dev/gardener-project/releases/gardener/extensions`. This swap from `/public` to `/releases` is done because pushing is not allowed to the `/public` repository and images created from the `Makefile` targets are meant to be pushed to a registry.

Fixes #

**Release note**:

```other operator
NONE
```


Dockerfile
```diff
@@ -31,7 +31,7 @@ COPY --from=builder /go/bin/gardener-extension-shoot-rsyslog-relp-admission /gar
 ENTRYPOINT ["/gardener-extension-shoot-rsyslog-relp-admission"]
 
 ############# rsyslog-relp-echo-server
-FROM alpine:3.18.4 AS shoot-rsyslog-relp-echo-server
+FROM europe-docker.pkg.dev/gardener-project/public/3rd/alpine:3.18.4 AS shoot-rsyslog-relp-echo-server
```
ialidzhikov, line 34: I think we don't do this for Dockerfiles. The whole idea of copying images to AR is to deal with dockerhhub rate limits and with registries that don't support image pulls over IPv6.
IMO, you could revert here.

ialidzhikov (commented):
Minor suggestion, rest lgtm.

ialidzhikov (approved):
/lgtm

PR #61, opened 2024-02-12
Clean up sed usage after ControllerRegistration generation

ialidzhikov:

/kind cleanup

The used values
```yaml
    lifecycle:
      reconcile: AfterKubeAPIServer
      delete: BeforeKubeAPIServer
      migrate: BeforeKubeAPIServer
```

should be the defaults already. See https://github.com/gardener/gardener/blob/cb27cb1913c879d6257a0576b07964f14c84f2ef/pkg/apis/core/v1beta1/types_controllerregistration.go#L78-L84
This `sed` usage is rather a leftover from copy-paste from an extension that needs custom lifecycle values.

N/A

N/A

**Release note**:

```other operator
NONE
```


plkokanov (approved):
/lgtm

I think I wanted to set them explicitly to the defaults for some reason. Maybe because I saw it was done like that in some other extensions. 
Anyway, you're right that it doesn't make too much sense to do it like that.

PR #76, opened 2024-03-06
Only use immutable tls secret

Kostov6:

/area quality
/kind enhancement

This PR forces referenced tls secret resource to be immutable by rejecting shoot config if mutable

Part of https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/issues/2

**Release note**:

```breaking user
When changing referenced TLS secret in `shoot.spec.resources[]` the user should provide only immutable secret
```

### DIFF
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -123,9 +123,11 @@ loggingRules:
 
 ### Securing the Communication to the Target Server with TLS
 
-The communication to the target server is not encrypted by default. To enable encryption, set the `.tls.enabled` field in the `shoot-rsyslog-relp` extension configuration to `true`. In this case, a secret which contains the TLS certificates used to establish the TLS connection to the server must be created in the same project namespace as your Shoot.
+The communication to the target server is not encrypted by default. To enable encryption, set the `.tls.enabled` field in the `shoot-rsyslog-relp` extension configuration to `true`. In this case, a immutable secret which contains the TLS certificates used to establish the TLS connection to the server must be created in the same project namespace as your Shoot.
 
 An example Secret is given below:
+
+> **Note:**  The secret should be immutable
 
 ```yaml
 kind: Secret
@@ -133,6 +135,7 @@ apiVersion: v1
 metadata:
   name: rsyslog-relp-tls-v1
   namespace: garden-foo
+immutable: true
 data:
   ca: |
     -----BEGIN BEGIN RSA PRIVATE KEY-----
--- a/docs/usage/configuration.md
+++ b/docs/usage/configuration.md
@@ -123,11 +123,11 @@ loggingRules:
 
 ### Securing the Communication to the Target Server with TLS
 
-The communication to the target server is not encrypted by default. To enable encryption, set the `.tls.enabled` field in the `shoot-rsyslog-relp` extension configuration to `true`. In this case, a immutable secret which contains the TLS certificates used to establish the TLS connection to the server must be created in the same project namespace as your Shoot.
+The communication to the target server is not encrypted by default. To enable encryption, set the `.tls.enabled` field in the `shoot-rsyslog-relp` extension configuration to `true`. In this case, an immutable secret which contains the TLS certificates used to establish the TLS connection to the server must be created in the same project namespace as your Shoot.
 
 An example Secret is given below:
 
-> **Note:**  The secret should be immutable
+> **Note:**  The secret must be immutable
 
 ```yaml
 kind: Secret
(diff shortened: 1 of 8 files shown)


plkokanov (changes requested):
Some comments inline.
Additionally, you should flag this as a breaking change.

pkg/utils/utils.go
```diff
@@ -34,6 +34,9 @@ func ValidateRsyslogRelpSecret(secret *corev1.Secret) error {
 	if _, ok := secret.Data[constants.RsyslogPrivateKeyKey]; !ok {
 		return fmt.Errorf("secret %s is missing %s value", key.String(), constants.RsyslogPrivateKeyKey)
 	}
+	if secret.Immutable == nil || !*secret.Immutable {
```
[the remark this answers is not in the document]
Kostov6: Shouldn't it be `if !ptr.Deref(secret.Immutable, false) {`?

pkg/utils/utils.go
```diff
@@ -34,6 +35,9 @@ func ValidateRsyslogRelpSecret(secret *corev1.Secret) error {
 	if _, ok := secret.Data[constants.RsyslogPrivateKeyKey]; !ok {
 		return fmt.Errorf("secret %s is missing %s value", key.String(), constants.RsyslogPrivateKeyKey)
 	}
+	if !ptr.Deref(secret.Immutable, false) {
```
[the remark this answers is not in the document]
Kostov6: Should the example secret in the document have a hash suffix?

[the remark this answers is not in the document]
ialidzhikov: It is already suffixed with v1 so I assume this v1 represents the revision of the Secret. At least this is also the naming pattern we use in the registry-cache docs when dealing with immutable Secrets: see the docs in https://github.com/gardener/gardener-extension-registry-cache/blob/9176c6572d582e0725024b1db30735d8c7f25cbc/docs/usage/registry-cache/upstream-credentials.md, you will find there `-v1` and `-v2` suffixes in the Secret names.

docs/usage/configuration.md
```diff
@@ -123,16 +123,19 @@ loggingRules:
 
 ### Securing the Communication to the Target Server with TLS
 
-The communication to the target server is not encrypted by default. To enable encryption, set the `.tls.enabled` field in the `shoot-rsyslog-relp` extension configuration to `true`. In this case, a secret which contains the TLS certificates used to establish the TLS connection to the server must be created in the same project namespace as your Shoot.
+The communication to the target server is not encrypted by default. To enable encryption, set the `.tls.enabled` field in the `shoot-rsyslog-relp` extension configuration to `true`. In this case, a immutable secret which contains the TLS certificates used to establish the TLS connection to the server must be created in the same project namespace as your Shoot.
```
plkokanov, line 126: nit:
```suggestion
The communication to the target server is not encrypted by default. To enable encryption, set the `.tls.enabled` field in the `shoot-rsyslog-relp` extension configuration to `true`. In this case, an immutable secret which contains the TLS certificates used to establish the TLS connection to the server must be created in the same project namespace as your Shoot.
```

docs/usage/configuration.md
```diff
@@ -123,16 +123,19 @@ loggingRules:
...
 
 An example Secret is given below:
 
+> **Note:**  The secret should be immutable
```
plkokanov, line 130:
```suggestion
> **Note:**  The secret must be immutable
```

pkg/utils/utils.go
```diff
@@ -34,6 +35,9 @@ func ValidateRsyslogRelpSecret(secret *corev1.Secret) error {
...
+		return fmt.Errorf("secret %s should be immutable", key.String())
```
plkokanov, line 39:
```suggestion
		return fmt.Errorf("secret %s must be immutable", key.String())
```

pkg/utils/utils_test.go
```diff
@@ -43,35 +43,41 @@ var _ = Describe("Utils", func() {
...
 		Entry(
 			"should return error if secret does not contain 'tls' data entry",
-			[]byte("caData"), []byte("crtData"), []byte("tlsData"), []byte("extraData"),
+			[]byte("caData"), []byte("crtData"), []byte("tlsData"), []byte("extraData"), true,
 			MatchError(ContainSubstring("secret foo/rsyslog-secret should have only three data entries")),
 		),
+		Entry(
+			"should return error if secret is mutable",
+			[]byte("caData"), []byte("crtData"), []byte("tlsData"), []byte("extraData"), false,
+			MatchError(ContainSubstring("secret foo/rsyslog-secret should be immutable")),
```
plkokanov, line 80:
```suggestion
			MatchError(ContainSubstring("secret foo/rsyslog-secret must be immutable")),
```

plkokanov (changes requested):
Few more minor nits related to grammar and proper terms. 
We could also include this PR in the new release.

PR #80, opened 2024-03-19
Always delete the config cleaner MR when the extension is deleted

plkokanov:

/kind bug

This PR fixes an issue in which the `extension-shoot-rsyslog-relp-configuration-cleaner` ManagedResource could prevent a shoot cluster from being deleted. The issue could happen like this:
1. Create a shoot cluster with the `shoot-rsyslog-relp` extension enabled
2. Disable the extension, which would cause the `extension-shoot-rsyslog-relp-configuration-cleaner` MR to get deployed
3. Force an error to occur during after the deployment of the MR is finished, but before it has become ready.
4. Delete the shoot cluster

When the reconciliation of the `shoot-rsyslog-relp` extension is restarted while the shoot cluster is being deleted, it will exit early [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/83bfaddbd8080a03818e42459b7c927a90938dfb/pkg/controller/lifecycle/actuator.go#L99-L104)
This means that the code which should trigger the deletion of the MR [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/83bfaddbd8080a03818e42459b7c927a90938dfb/pkg/controller/lifecycle/actuator.go#L175-L177) never gets executed.

In the end the MR is orphaned and blocks shoot deletion.

---

To fix the issue we now only skip the deployment and waiting for the `extension-shoot-rsyslog-relp-configuration-cleaner` MR to become ready if the shoot is being deleted. This MR is not necessary in that case as the nodes will get deleted anyway. Additionally, due to various problems, waiting for the MR to become ready might block shoot deletion.
The deletion of the MR is done in all cases now.

None

**Release note**:

```bugfix operator
Fixed an issue where the `extension-shoot-rsyslog-relp-configuration-cleaner` ManagedResource could block Shoot deletion if the `shoot-rsyslog-relp` was disabled before the Shoot deletion was triggered, and disabling the extension failed while trying to deploy the said ManagedResource and wait for it to become ready. 
```

### DIFF
--- a/pkg/controller/lifecycle/actuator.go
+++ b/pkg/controller/lifecycle/actuator.go
@@ -161,9 +161,12 @@ func cleanRsyslogRelpConfiguration(ctx context.Context, cluster *extensionscontr
 	}
 	cleaner := rsyslogrelpconfigcleaner.New(client, namespace, values)
 
-	// If the Shoot is in deletion, then there is no need to deploy the component to clean up
-	// the rsyslog configuration from Nodes. However, we should still try to destroy the
-	// configuration cleaner component, in case it was unsuccessfully deployed in a previous reconciliation
+	// If the Shoot is in deletion, then there is no need to deploy the component to clean up the rsyslog
+	// configuration from Nodes. The Shoot deletion flow ensures that the Worker is deleted before
+	// the Extension deletion. Hence, there are no Nodes, no need to deploy the configuration cleaner component.
+	//
+	// However, we should still try to destroy the
+	// configuration cleaner component, in case it failed to be cleaned up in a previous reconciliation
 	// where the extension was deleted before the shoot deletion was triggered.
 	if cluster.Shoot.DeletionTimestamp == nil {
 		if err := component.OpWait(cleaner).Deploy(ctx); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/lifecycle/actuator.go
```diff
@@ -168,8 +161,14 @@ func cleanRsyslogRelpConfiguration(ctx context.Context, cluster *extensionscontr
 	}
 	cleaner := rsyslogrelpconfigcleaner.New(client, namespace, values)
 
-	if err := component.OpWait(cleaner).Deploy(ctx); err != nil {
-		return fmt.Errorf("failed to deploy the rsyslog relp configuration cleaner component: %w", err)
+	// If the Shoot is in deletion, then there is no need to deploy the component to clean up
+	// the rsyslog configuration from Nodes. However, we should still try to destroy the
+	// configuration cleaner component, in case it was unsuccessfully deployed in a previous reconciliation
+	// where the extension was deleted before the shoot deletion was triggered.
```
ialidzhikov, line 167: Cosmetic nit: I would still preserve the 2 sentences that explain why we don't need to deploy the configuration cleaner component on Shoot deletion. This is obvious for us but can be confusing for an external reader.

```suggestion
	// If the Shoot is in deletion, then there is no need to deploy the component to clean up the rsyslog
	// configuration from Nodes. The Shoot deletion flow ensures that the Worker is deleted before
	// the Extension deletion. Hence, there are no Nodes, no need to deploy the configuration cleaner component.
	//
	// However, we should still try to destroy the
	// configuration cleaner component, in case it failed to be cleaned up in a previous reconciliation
	// where the extension was deleted before the shoot deletion was triggered.
```

PR #82, opened 2024-03-23
Increase timeouts for `testmachinery` tests

plkokanov:

/area testing
/kind bug

This PR fixes an issue in the `testmachinery` by increasing their timeout and also adding proper timeout and polling interval arguments to `Eventually` and `Consistnetly` calls.
Additionally, the expected message log that is retrieved from the `rsyslog-relp-echo-server` is now also matched against the hostname of the node from which the message originated.

Part of #2 

I added an additional option to configure the `tlsLib` setting of the `shoot-rsyslog-relp` extension for local testing purposes. However, I decided to leave it for the future in case we want to create an additional test with `gnutls`. Currently only `openssl` is tested. The reason for that is that certificates have to be generated with an appropriate `cert-tool` as certificates for `openssl` are not compatible with `gnutls` and vice-versa.

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/common/verifier.go
+++ b/test/common/verifier.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/gardener/gardener/pkg/client/kubernetes"
 	"github.com/gardener/gardener/test/framework"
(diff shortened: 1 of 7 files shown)


test/common/verifier.go
```diff
@@ -49,6 +54,27 @@ func NewVerifier(log logr.Logger,
...
+	v.consistentlyArgs = args
+	return v
+}
+
+// WithEventuallyArgs specifies the arguments to pass to
+// `Eventually` calls made by the Verifier.
+func (v *Verifier) WithEventuallyArgs(args ...interface{}) *Verifier {
+	v.eventuallyArgs = args
+	return v
+}
```
ialidzhikov, line 69: IIUC, the motivation for this change is to have different timeout and poll interval for e2e and TM tests. 
However, we found these options confusing as it is not very clear to the reader for which assertions these timeouts/poll intervals are used. Tomorrow, you may need to add another assertion in the test and how we will make the diff between the eventuallyArgs used for assertion foo and the eventuallyArgs used for assertion bar.

We would rather suggest to hard-code the higher timeouts/poll intervals. The e2e tests will be simply faster and won't need to whole time. Or, is there are any strong reason to have different values for the different types of tests?

test/common/verifier.go
```diff
@@ -140,5 +166,19 @@ func (v *Verifier) generateAndGetLogs(ctx context.Context, programName, severity
 }
 
 func (v *Verifier) constructRegex(programName, logMessage string) string {
-	return fmt.Sprintf(` %s %s %s .* %s\[\d+\]: .* %s`, v.projectName, v.shootName, v.shootUID, programName, logMessage)
+	expectedNodeHostName := v.nodeName
+	switch {
+	case v.providerType == "aws":
+		// On aws the the nodes ara named by adding the regional domain name after the instance, e.g.:
```
ialidzhikov, line 172:
```suggestion
		// On aws the nodes are named by adding the regional domain name after the instance, e.g.:
```

test/common/verifier.go
```diff
@@ -140,5 +166,19 @@ func (v *Verifier) generateAndGetLogs(ctx context.Context, programName, severity
...
+		// `ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`. However the rsyslog `hostname` property only returns the
+		// instance name - `ip-xxx-xxx-xxx-xxx`.
+		expectedNodeHostName = strings.SplitN(v.nodeName, ".", 2)[0]
+	case v.providerType == "alicloud":
+		// On alicloud the name of a node is made of lower case characters, e.g. `izgw846obiag360olq8sdaz`.
+		// However, the rsyslog `hostname` property can also contain upper case characters, e.g. `iZgw846obiag360olq8sdaZ`.
+		// This is why we use the (?i:...) - to turn on case-insensitive mode for the hostname matching.
+		expectedNodeHostName = "(?i:" + expectedNodeHostName + ")"
+	}
```
ialidzhikov, line 181: A cosmetic suggestion that is up to you whether to incorporate or not:
```suggestion
	var expectedNodeHostName string

	switch {
	case v.providerType == "aws":
		// On aws the the nodes ara named by adding the regional domain name after the instance, e.g.:
		// `ip-xxx-xxx-xxx-xxx.ec2.<region>.internal`. However the rsyslog `hostname` property only returns the
		// instance name - `ip-xxx-xxx-xxx-xxx`.
		expectedNodeHostName = strings.SplitN(v.nodeName, ".", 2)[0]
	case v.providerType == "alicloud":
		// On alicloud the name of a node is made of lower case characters, e.g. `izgw846obiag360olq8sdaz`.
		// However, the rsyslog `hostname` property can also contain upper case characters, e.g. `iZgw846obiag360olq8sdaZ`.
		// This is why we use the (?i:...) - to turn on case-insensitive mode for the hostname matching.
		expectedNodeHostName = "(?i:" + expectedNodeHostName + ")"
	default:
		expectedNodeHostName = v.nodeName
	}
```

ialidzhikov (commented):
We had a look together with @dimitar-kostadinov and @Kostov6. Minor suggestions, otherwise lgtm.

plkokanov:
Thanks for the review! I have addressed your comments, PTAL

ialidzhikov (approved):
/lgtm

PR #83, opened 2024-04-01
Fix tm tests for alicloud provider

plkokanov:

/area testing
/kind bug

This PR fixes an issue with the TM tests on the alicloud provider that was introduced after the recommendations from the PR review [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/pull/82#discussion_r1538917133) were applied. When the provider is alicloud the `v.nodeName` field should be used to initialise the `expectedHostName` variable instead of reusing the `expectedHostName` variable itself which is defaulted to an empty string at that point.

Fixes #

/cc @ialidzhikov @Kostov6 

**Release note**:

```other operator
NONE
```


ialidzhikov (approved):
/lgtm

PR #84, opened 2024-04-04
Removes comments at start of audit rules files as they do not work on all OSes

plkokanov:

/kind bug

This PR removes the license comments at the start of audit rules files as they do not get properly parsed when `augenrules --load` is invoked 
Additionally it re-adds the name of the `omrelp` action which was regressed in #41 

Fixes #

**Release note**:

```other operator
NONE
```


Kostov6 (commented):
/lgtm

PR #85, opened 2024-04-08
Adds empty lines to end of audit rule files

plkokanov:

/kind bug

This PR adds empty lines to the end of audit rule files and partially reverts #84 

The actual issue in #84 was that comments in audit rule files are expected to be defined with `##` instead of only `#`.
Additionally, there was one more issue where the audit rule files need to have an extra empty line at the end. Otherwise they get merged incorrectly into `/etc/audit/audit.rules` when calling `augenrules --load` 

Fixes #

Tested on a shoot cluster using `gardenlinux` and `suse-chost`

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -24,8 +24,6 @@ Files:
   example/controller-registration.yaml
   hack/api-reference/*.json
   hack/cherry-pick-pull.sh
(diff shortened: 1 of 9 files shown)


Kostov6 (commented):
/lgtm

PR #87, opened 2024-04-10
Add dependabot config

plkokanov:

/area dev-productivity
/kind enhancement

Adds `dependabot.yaml` config for automatic upgrades of `gardener/gardener` and `golang` versions.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.github/workflows/vendor_gardener.yaml
+++ b/.github/workflows/vendor_gardener.yaml
@@ -1,0 +1,31 @@
+name: Vendor Gardener
+on:
+  push:
+    branches:
+    - dependabot/go_modules/github.com/gardener/**
+permissions: write-all
+jobs:
+  run:
+    name: Run make tidy
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-go@v3
+      with:
+        go-version-file: go.mod
+
+    - name: Make tidy
+      run: make tidy
+    - name: Commit changes
+      run: |
+        # Exit early if there is nothing to commit. This can happen if someone pushes to the dependabot's PR (for example has to adapt to a breaking change).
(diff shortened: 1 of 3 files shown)


.github/workflows/vendor_gardener.yaml
```diff
@@ -0,0 +1,31 @@
+name: Vendor Gardener
```
ialidzhikov, line 1: In https://github.com/gardener/gardener-extension-registry-cache/pull/147/commits/f683d23a862bd1473ebc30ddf118213247ccfb99, I dropped this Github action with the vendor dir removal. I think the only use of this Github action after the vendor dir removal would be to update the 2 script that are currently coming from gardener (component_descriptor and prepare_release). I decided to manually maintain these 2 scripts instead of having such Github action.

No strong opinion, I just wanted to share the approach I decided to follow for the registry-cache repo.

plkokanov: Yep I noticed that, but since I sometimes tend to forget such stuff, I decided to leave the action :)

ialidzhikov, line 1: Maybe you could rename the file and the name field to something other than vendoring? (as we no longer have the vendor dir or make target named vendor)

plkokanov: wdyt about something like "Sync Gardener Script Dependencies"

ialidzhikov: No strong opinion. The above suggestion is ok. Alternatively, you can use something around make tidy as at the end the github action does this. Whatever you prefer.

ialidzhikov (approved):
/lgtm

ialidzhikov:
/hold
to unblock tide - this PR has changes under .github and tide cannot merge such PRs

PR #90, opened 2024-04-12
Make monitoring `ConfigMap` mutable and explicitly delete old immutable monitoring `ConfigMap`s

### LINKED ISSUE #89: The `shoot-rsyslog-relp-prometheus-*` configmap used for monitoring config could cause `prometheus` to `CrashLoopBackOff` after version updates
**How to categorize this issue?**

/area monitoring
/kind bug

**What happened**:

When the `shoot-rsyslog-relp` extension controller is updated, the following sequence of events could occur if the update contains changes to the monitoring config:

1. A new release of `shoot-rsyslog-relp` extension which contains changes to the monitoring config is deployed to a landscape.
2. When the `shoot-rsyslog-relp` extension resource is reconciled as part of shoot reconciliation, it will deploy a new monitoring configmap containing the updated monitoring config
3. We then expect that GCM's garbage collection will delete the old monitoring configmap.
4. The shoot reconciliation should then pick all monitoring configmaps deployed by extensions and merge them into the configmap which is used to configure prometheus.
5. However, there is a chance that (3.) does not happen quick enough, so there will be two immutable configmaps for the `shoot-rsyslog-relp` extension. Shoot reconciliation can pick both, and add them to the prometheus configmap leading to the following error:
```
ts=2024-04-12T01:21:58.620Z caller=main.go:521 level=error msg="Error loading config (--config.file=/etc/prometheus/config/prometheus.yaml)" file=/etc/prometheus/config/prometheus.yaml err="parsing YAML file /etc/prometheus/config/prometheus.yaml: found multiple scrape configs with job name \"rsyslog-metrics\""
```

**What you expected to happen**:

Prometheus to not enter a `CrashLoopBackOff`

**How to reproduce it (as minimally and precisely as possible)**:

This is hard to reproduce as it occurs due to a race condition. However, here are the general steps:
1. Prepare a garden cluster with the `rsyslog-relp-extension` and create a shoot cluster that uses the extension by following the [getting started guide](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/main/docs/development/getting-started.md)
2. Wait for the shoot cluster to be reconciled successfully
3. Make a change to the monitoring config here: https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/443d47de590cd6fdaf44b35ddf97e955700f4d5e/pkg/controller/lifecycle/monitoring_config.go#L40-L84
4. Rerun `make extension-up` and check whether the `prometheus-0` pod in the shoot's control plane enters `CrashLoopBackOff`

**Anything else we need to know?**:

**Environment**:

- Gardener version (if relevant):
- Extension version:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- Others:

plkokanov:

/area monitoring
/kind bug

This PR makes the `ConfigMap` used for the monitoring configurations deployed by the `shoot-rsyslog-relp` extension `mutable`. It was incorrectly marked as `immutable` which could lead to multiple monitoring `ConfigMap`s existing at the same time and getting aggregated into the `ConfigMap` used to configure prometheus. This would lead to the following error:
```
ts=2024-04-12T01:21:58.620Z caller=main.go:521 level=error msg="Error loading config (--config.file=/etc/prometheus/config/prometheus.yaml)" file=/etc/prometheus/config/prometheus.yaml err="parsing YAML file /etc/prometheus/config/prometheus.yaml: found multiple scrape configs with job name \"rsyslog-metrics\""
```

As a transition step, the PR also includes code that will delete the old immutable `ConfigMap`s immediately to cover cases where they would not get deleted by the `gardener-resource-manager`'s garbage collection controller quickly enough.

Fixes #89 

**Release note**:

```other operator
The `ConfigMap` deployed for the monitoring configuration of the `shoot-rsyslog-relp` extension  in `Shoot` control planes is no longer `immutable`. This fixes an issue that could cause `prometheus-0` pods to get stuck in `CrashLoopBackOff`.
```

### DIFF
--- a/pkg/controller/lifecycle/monitoring.go
+++ b/pkg/controller/lifecycle/monitoring.go
@@ -12,16 +12,26 @@ import (
 	"github.com/gardener/gardener/pkg/controllerutils"
 	"github.com/gardener/gardener/pkg/resourcemanager/controller/garbagecollector/references"
 	"github.com/gardener/gardener/pkg/utils"
-	kubernetesutils "github.com/gardener/gardener/pkg/utils/kubernetes"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/gardener/gardener-extension-shoot-rsyslog-relp/pkg/constants"
 )
 
-func deployMonitoringConfig(ctx context.Context, client client.Client, namespace string) error {
+func deployMonitoringConfig(ctx context.Context, c client.Client, namespace string) error {
+	// TODO(plkokanov): remove this in a future release.
+	// Refer to https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/issues/89 for more details.
+	if err := c.DeleteAllOf(ctx, &corev1.ConfigMap{},
+		client.InNamespace(namespace),
+		client.MatchingLabels{
+			"component": constants.ServiceName,
+			"extensions.gardener.cloud/configuration": "monitoring",
+			references.LabelKeyGarbageCollectable:     references.LabelValueGarbageCollectable,
+		}); err != nil {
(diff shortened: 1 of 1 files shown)


ialidzhikov (approved):
/lgtm

🚀

PR #92, opened 2024-04-15
Do not wait for deletion of rsyslog MR if shoot is hibernated

plkokanov:

/kind bug

This PR makes it so that the clean up code added in #41 to delete the no-longer-used `extension-shoot-rsyslog-relp-shoot` `ManagedResource` does not wait for the deletion if the `Shoot` is hibernated. This is necessary to prevent errors like:
```
'Error reconciling Extension: retry failed with context deadline
      exceeded, last error: resource <shoot-controlplane-namespace>/extension-shoot-rsyslog-relp-shoot
      still exists'
```

Fixes #

Note that we have to be certain that the `Daemonset` deployed by the MR is removed so that it does not interfere with changes made by the OSC once the workers are created later in the reconciliation flow when the shoot is woken up. Thus this clean up code has to be kept for longer.

/cc @shafeeqes 

**Release note**:

```other operator
The reconciliation of the `shoot-rsyslog-relp` extension no longer waits for the `extension-shoot-rsyslog-relp-shoot` `MangedResource` to be deleted during reconciliations, if the `Shoot` cluster is hibernated. The wait will still be executed when the `Shoot` is woken up to ensure that the resources deployed in the `Shoot` are removed. 
```

### DIFF
--- a/pkg/controller/lifecycle/actuator.go
+++ b/pkg/controller/lifecycle/actuator.go
@@ -57,18 +57,32 @@ type actuator struct {
 func (a *actuator) Reconcile(ctx context.Context, _ logr.Logger, ex *extensionsv1alpha1.Extension) error {
 	namespace := ex.GetNamespace()
 
+	if err := deployMonitoringConfig(ctx, a.client, namespace); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


shafeeqes (commented):
Thank you!
/lgtm

PR #104, opened 2024-05-16
Avoid audit log duplication when `system-journald-audit` socket is enabled

plkokanov:

/kind bug

This PR fixes an issue that caused audit logs to be duplicated when the `system-journald-audit` socket is enabled on the OS. 
Generally, the extension enables the `syslog` plugin for `auditd` in order to forward audit logs to journald on all supported OSes. However, on gardenlinux `system-journald-audit` socket is also enabled, meaning that journald also fetches audit logs from there. This caused audit event messages to be duplicated in journald.

The PR makes it so that the `system-journal-audit` socket is disabled and stopped when the extension is used to align how audit logs are gathered across multiple OSes.
If the extension is disabled, the `system-journal-audit` socket is enabled and started again.

Part of #2

**Release note**:

```other operator
Fixed an issue that caused audit logs to be duplicated in journald if the `system-journald-audit` socket was enabled. Now if the `system-journald-audit` socket exists on the node, it is disabled and stopped when this extension is used.
```

### DIFF
--- a/pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
+++ b/pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
@@ -13,17 +13,40 @@ function configure_auditd() {
     mv /etc/audit/rules.d /etc/audit/rules.d.original
   fi
 
+  restart_auditd=false
+
   if [[ ! -d /etc/audit/rules.d ]]; then
     mkdir -p /etc/audit/rules.d
   fi
   if ! diff -rq /var/lib/rsyslog-relp-configurator/audit/rules.d /etc/audit/rules.d ; then
     rm -rf /etc/audit/rules.d/*
     cp -L /var/lib/rsyslog-relp-configurator/audit/rules.d/* /etc/audit/rules.d/
-    if [[ -f /etc/audit/plugins.d/syslog.conf ]]; then
-      sed -i 's/no/yes/g' /etc/audit/plugins.d/syslog.conf
-    fi
     augenrules --load
-    systemctl restart auditd
+    restart_auditd=true
+  fi
+
+  if [[ -f /etc/audit/plugins.d/syslog.conf ]] && grep -Fxq "active = no" "/etc/audit/plugins.d/syslog.conf" ; then
+    sed -i 's/no/yes/g' /etc/audit/plugins.d/syslog.conf
+    restart_auditd=true
+  fi
+
+  if ! systemctl is-active --quiet auditd.service ; then
+    # Ensure that the auditd service is running.
+    systemctl start auditd.service
+  elif [ "${restart_auditd}" = true ]; then
+    systemctl restart auditd.service
+  fi
+
+  # If the `systemd-journald-audit.socket` socket exists and is enabled, then journald also fetches audit logs from it.
+  # To avoid duplication we disable it and only rely on the syslog audit plugin.
+  if systemctl list-unit-files systemd-journald-audit.socket > /dev/null ; then
+    if systemctl is-enabled --quiet systemd-journald-audit.socket ; then
+      systemctl disable systemd-journald-audit.socket
+    fi
+    if systemctl is-active --quiet systemd-journald-audit.socket ; then
+      systemctl stop systemd-journald-audit.socket
+      systemctl restart systemd-journald
+    fi
   fi
 }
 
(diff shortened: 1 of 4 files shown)


pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
```diff
@@ -13,17 +13,40 @@ function configure_auditd() {
...
+
+  if [[ -f /etc/audit/plugins.d/syslog.conf ]] && grep -Fxq "active = no" "/etc/audit/plugins.d/syslog.conf" ; then
+    sed -i 's/no/yes/g' /etc/audit/plugins.d/syslog.conf
+    restart_auditd=true
+  fi
+
+  if ! systemctl is-active --quiet auditd.service ; then
+    # Ensure that the auditd service is running.
+    systemctl start auditd.service
+  elif [ "${restart_auditd}" = true ]; then
```
Kostov6, line 36:
```suggestion
  elif [ "$restart_auditd" = true ]; then
```
Could we drop the {}? I think the expression is simple enough for them to not carry any benefit

pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
```diff
@@ -13,17 +13,40 @@ function configure_auditd() {
...
-    if [[ -f /etc/audit/plugins.d/syslog.conf ]]; then
-      sed -i 's/no/yes/g' /etc/audit/plugins.d/syslog.conf
-    fi
     augenrules --load
-    systemctl restart auditd
+    restart_auditd=true
+  fi
+
+  if [[ -f /etc/audit/plugins.d/syslog.conf ]] && grep -Fxq "active = no" "/etc/audit/plugins.d/syslog.conf" ; then
+    sed -i 's/no/yes/g' /etc/audit/plugins.d/syslog.conf
```
Kostov6, line 29:
```suggestion
    sed -i 's/active = no/active = yes/' /etc/audit/plugins.d/syslog.conf
```
I think this could have unexpected behaviour if the the file has `<another_field> = no`. Also the g option should not be necessary. Also if it is expected for the user to be able to potentially edit this file do we have the guarantee that there would be spaces between "active" and "no"(ex `active=no`)?

pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
Kostov6, line 36:
```suggestion
  elif [ "$restart_auditd" = true ]; then
```
Could we drop the {}? I think the expression is simple enough for them to not carry any benefit

Kostov6 (commented):
Minor nits and a question from my side. Otherwise lgtm

pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
plkokanov: That is the only option containing "no" or "yes" as per the specification. However, this could change in the future so I'll adapt the script.

pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
plkokanov: You're right that it is not required in this case, however I've left them for consistency reasons - this is something we try to do in all our bash scripts.

plkokanov:
@Kostov6 could you take another look so that we can get this in?

pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
```diff
@@ -25,9 +25,11 @@ function configure_auditd() {
...
   fi
 
-  if [[ -f /etc/audit/plugins.d/syslog.conf ]] && grep -Fxq "active = no" "/etc/audit/plugins.d/syslog.conf" ; then
-    sed -i 's/no/yes/g' /etc/audit/plugins.d/syslog.conf
-    restart_auditd=true
+  if [[ -f /etc/audit/plugins.d/syslog.conf ]] && \
+      grep -m 1 -qie  "^active\\>" "/etc/audit/plugins.d/syslog.conf" && \
+      ! grep -m 1 -qie "^active\\> = yes" "/etc/audit/plugins.d/syslog.conf" ; then
+    sed -i "s/^active\\>.*/active = yes/gi" /etc/audit/plugins.d/syslog.conf
+    export restart_auditd=true
```
Kostov6, line 32: By what process is `restart_auditd` consumed? Aren't env vars by convention upper case?

Kostov6 (commented):
Just a small question from my side. Otherwise lgtm

pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh
plkokanov: Nice catch, not sure why I added the `export` keyword here

plkokanov: I've removed it now. PTAL.

PR #106, opened 2024-05-21
Restart rsyslog when tls certificates change

plkokanov:

/kind bug

Previously rsyslog was not restarted when the tls certificates were changed. This seems to cause issues as the new certificates aren't properly loaded by rsyslog when it tries to execute the action to send logs over the relp channel.

WIth this PR the tls certificates are copied from the directory where they are created by the `gardener-node-agent` to `/etc/ssl/rsyslog` and loaded into rsyslog from there. If the certificates downloaded by `gardener-node-agent` are changed, the old ones are removed from `/etc/ssl/rsyslog`, new ones copied over from their original directory and the rsyslog service is restarted

Additionally, a small change for clarity was made by moving the `input(type="imuxsock" Socket="/run/systemd/journal/syslog")` input a bit up in the rsyslog configuration file.

Part of #2

**Release note**:

```other operator
If the certificates used for the `rsyslog-relp` tls connection are changed, the `rsyslog` service on the nodes is restarted so that it can properly load the new certificates.
```


ialidzhikov (commented):
/lgtm

PR #111, opened 2024-05-30
Increase permissions of `/etc/ssl/rsyslog` dir

plkokanov:

/kind bug

It seems that when rsyslog is started without `cap_dac_override` as is the case on some OSes, it cannot read from the `/etc/ssl/rsyslog` directory when tls is enabled even if the directory has 0600 permissions. With this PR we will create the directory with default (0755) permissions to avoid this problem. Certificates inside it still have `0600` permissions which does not cause any problems.

Fixes #

**Release note**:

```other operator
The directory where the tls certificates are copied on the node - `/etc/ssl/rsyslog`, is now created with default (`0755`) permissions so that it can be read by an `rsyslog` process that is started without `cap_dac_override` capability. 
```


Kostov6 (commented):
```
  -m, --mode=MODE   set file mode (as in chmod), not a=rwx - umask
  -p, --parents     no error if existing, make parent directories as needed,
                    with their file modes unaffected by any -m option.
```
Do we need the -p flag if `mkdir -m 0600 <>` was running successfully?

plkokanov:
I added it as a precaution if the directory structure changes in the future. And I guess it is safer if some os does not have `/etc/ssl` dir.

Kostov6 (commented):
/lgtm

PR #115, opened 2024-06-04
Adds a queue for rsyslog relp messages

plkokanov:

/kind enhancement

This PR adds a `linkedlist` in-memory queue for the `rsyslog-relp` action. The queue will be used to store messages if the relp server cannot be reached for whatever reason. Additionally, a disk queue is also added which allows to store messages on disk under `/var/log/rsyslog`. This disk queue is used when the in-memory queue is exhausted or when rsyslog is restarted to save the current messages that were added to the in-memory queue.

Previously, only the input queue was used which was capped at 50000 messages. Using the larger rsyslog-relp action queue does not increase the consumed memory too much and it allows for longer downtimes of the rsyslog-relp server.

More information on rsyslog queues can be found in https://www.rsyslog.com/doc/concepts/queues.html and https://www.rsyslog.com/doc/rainerscript/queue_parameters.html

Part of #2

**Release note**:

```other operator
The `rsyslog-relp` action which is used to forward logs to a RELP server now uses a separate in-memory queue of 100000 messages. Additionally, it also uses a disk queue of max 48 MiB which is used to store messages after the in-memory queue is exhausted or to save the current messages in the in-memory queue when the `rsyslog` service is restarted.
```


plkokanov:
ping @Kostov6

pkg/component/rsyslogrelpconfigcleaner/rsyslog_relp_config_cleaner.go
```diff
@@ -177,6 +177,10 @@ func computeCommand() []string {
   chroot /host /bin/bash -c 'systemctl disable rsyslog-configurator; systemctl stop rsyslog-configurator; rm -f /etc/systemd/system/rsyslog-configurator.service'
 fi
 
+if [[ -d /host/var/log/rsyslog ]]; then
+  rm -rf /host/var/log/rsyslog
+fi
```
Kostov6, line 182:
```suggestion
rm -rf /host/var/log/rsyslog
```
```
$ rm --help | grep -e '--force'
  -f, --force           ignore nonexistent files and arguments, never prompt
```
Same for `test/integration/controller/lifecycle/lifecycle_test.go`

pkg/webhook/operatingsystemconfig/resources/templates/scripts/configure-rsyslog.tpl.sh
```diff
@@ -33,6 +33,10 @@ function configure_rsyslog() {
     systemctl enable rsyslog.service
   fi
 
+  if [[ ! -d {{ .rsyslogRelpQueueSpoolDir }} ]]; then
+    mkdir -p {{ .rsyslogRelpQueueSpoolDir }}
+  fi
```
Kostov6, line 38:
```suggestion
mkdir -p {{ .rsyslogRelpQueueSpoolDir }}
```
```
$ mkdir --help | grep -e "-p"
  -p, --parents     no error if existing, make parent directories as needed,
```
Same for `pkg/webhook/operatingsystemconfig/testdata/configure-rsyslog.sh`

Kostov6 (commented):
Minor nits, otherwise lgtm

plkokanov:
Thanks for the review, those `if`s were a leftover from a previous implementation that I was trying out.

pkg/component/rsyslogrelpconfigcleaner/rsyslog_relp_config_cleaner.go
plkokanov: This was to avoid the overhead of calling `rm` or `mkdir` multiple times since the script is run relatively often.

Kostov6:
```
$ time rm -rf foo

real	0m0.007s
user	0m0.002s
sys	0m0.003s

$ time mkdir -p go

real	0m0.018s
user	0m0.002s
sys	0m0.003s
```

We are in a nitpick territory between:
+2 lines and +0.02s speedup VS -2 lines and -0.02s speedup

So I'll give a lgtm and leave it up to you if you want to add my proposal

PR #127, opened 2024-06-27
Improve testmachinery tests and add checks for forwarding of audit logs

plkokanov:

/area testing
/kind enhancement

This PR tries to improve mainly the testmachinery tests by reducing the `exec` calls that were previously performed - instead of using a separate `exec` to generating each log line, now one `exec` will be used to generate all logs.

Additionally, forwarding of audit logs is now also checked. If instructed to, the test will create a file under the `/etc` directory. As per the audit rules this should generate an audit event of the file creation. The audit event contains the name of the file and the operation that created it. During the verification we look for the name of the file in the log lines retrieved from the echo server.

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/.test-defs/TestSuiteShootRsyslogRelpBetaSerial.yaml
+++ b/.test-defs/TestSuiteShootRsyslogRelpBetaSerial.yaml
@@ -6,7 +6,7 @@ spec:
   owner: gardener-oq@listserv.sap.com
   description: shoot-rsyslog-relp extension test suite that includes all serial beta tests
 
-  activeDeadlineSeconds: 10200
+  activeDeadlineSeconds: 12300
   labels: ["shoot", "beta"]
   behavior:
   - serial
--- a/.test-defs/TestSuiteShootRsyslogRelpBetaSerial.yaml
+++ b/.test-defs/TestSuiteShootRsyslogRelpBetaSerial.yaml
@@ -21,5 +21,6 @@ spec:
       -shoot-name=$SHOOT_NAME
       -ginkgo.focus="\[BETA\].*\[SERIAL\]"
       -ginkgo.skip="\[DISRUPTIVE\]"
+      -ginkgo.timeout=2h
 
   image: golang:1.22.5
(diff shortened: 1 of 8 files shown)


test/common/verifier.go
```diff
@@ -160,5 +214,18 @@ func (v *Verifier) constructRegex(programName, logMessage string) string {
...
+		matchRegexp := MatchRegexp(fmt.Sprintf(`%s %s %s %s \d+ %s\[\d+\]: .* %s`, v.projectName, v.shootName, v.shootUID, expectedNodeHostName, logEntry.program, logEntry.message))
+		if logEntry.shouldBeForwarded {
+			forwardedLogMatchers = append(forwardedLogMatchers, matchRegexp)
+		} else {
+			notForwardedLogMatchers = append(notForwardedLogMatchers, matchRegexp)
+		}
+	}
+
+	if v.testAuditLogging {
+		forwardedLogMatchers = append(forwardedLogMatchers, MatchRegexp(fileForAuditEvent))
```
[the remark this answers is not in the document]
plkokanov: Yep, `Contains` would be better here as we just want to make sure that the log line contains the file name. Thanks for the suggestion!

.test-defs/TestSuiteShootRsyslogRelpBetaSerial.yaml
```diff
@@ -21,5 +21,6 @@ spec:
       -shoot-name=$SHOOT_NAME
       -ginkgo.focus="\[BETA\].*\[SERIAL\]"
       -ginkgo.skip="\[DISRUPTIVE\]"
+      -ginkgo.timeout=2h
```
ialidzhikov, line 24: I would align this timeout with `activeDeadlineSeconds` as we were maintaining in activeDeadlineSeconds the sum of all test timeouts.

```suggestion
      -ginkgo.timeout=12000s
```

We have 2 tests using 50min as timeout, and 1 test now using 75min, so these 2h might not be enough in all cases.

plkokanov: Yeah, let's keep it consistent. Actually 13200 should be the correct value for `activeDeadlineSeconds`. I guess 12300 was a typo or miscalculation. I've also set the `ginkgo.timeout` to that value as well.

ialidzhikov (approved):

PR #128, opened 2024-06-28
Ignore errors when loading audit rules

plkokanov:

/kind enhancement

Audit rules are loaded sequentially and once an error is encountered, no further rules are loaded. This PR makes it so that errors are ignored when loading the predefined linux audit rules which are part of this extension. Errors generally come from kernel versions that do not support a particular system call part of the rules. This allows all rules to be loaded with the exception of the erroneous ones.
This change concretely targets the case where the `mount_setattr` system call is not available in linux kernel versions <= `5.12` (it is the only system call which causes this problem from the predefined rules). This is why it is now also extracted in a separate rule.

Part of #2

As part of the task which exposes audit rules so that they can be configured by users, I will add monitoring/alerts for errors/warnings that occur when loading them. 
Whether or not to set the `-i` option (so that errors are ignored and reported as warnings) will also be left to the users to decide.

**Release note**:

```other operator
Errors that can occur when loading audit rules are now ignored and reported as warnings. This allows all correct audit rules to be loaded.
```

### DIFF
--- a/pkg/webhook/operatingsystemconfig/resources/auditrules/11-privileged-special.rules
+++ b/pkg/webhook/operatingsystemconfig/resources/auditrules/11-privileged-special.rules
@@ -6,4 +6,8 @@
 ## The original file was moved to /etc/audit/rules.d.original
 
 -a exit,always -F arch=b64 -S mount -S umount2 -S mknod -S mknodat -S chroot -F auid!=-1 -F key=privileged_special
+
+## The mount_setattr kernel function is available only linux kernel versions >= 5.12.
+## It is specified in a separate rule so that the rule for the system
+## calls above can be safely loaded on kernel versions < 5.12.
 -a exit,always -F arch=b64 -S mount_setattr -F auid!=-1 -F key=privileged_special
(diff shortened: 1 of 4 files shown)


pkg/webhook/operatingsystemconfig/resources/auditrules/11-privileged-special.rules
```diff
@@ -5,4 +5,5 @@
 ## This file is managed by the shoot-rsyslog-relp extension
 ## The original file was moved to /etc/audit/rules.d.original
 
--a exit,always -F arch=b64 -S mount -S mount_setattr -S umount2 -S mknod -S mknodat -S chroot -F auid!=-1 -F key=privileged_special
+-a exit,always -F arch=b64 -S mount -S umount2 -S mknod -S mknodat -S chroot -F auid!=-1 -F key=privileged_special
+-a exit,always -F arch=b64 -S mount_setattr -F auid!=-1 -F key=privileged_special
```
ialidzhikov, line 9: Does it make sense to add a comment that mount_setattr is availble only in kernel versions >= 5.12?

plkokanov: Yep, makes perfect sense. I've added the comment now

ialidzhikov (approved):
/lgtm

andrerun (approved):

PR #136, opened 2024-07-10
Fix remote extension setup

plkokanov:

/area dev-productivity
/kind regression

This PR fixes the remote extension setup.

With https://github.com/gardener/gardener/pull/9782 the name with which gardener images are built with skaffold was changed and consequently the kyverno policies were changed to reflect that: https://github.com/gardener/gardener/pull/9782/files#diff-98574cc407c84776352d2e2fdea6d82ad1e194fa7a5961c14cdfe2eb2bea1271
This caused the `shoot-rsyslog-relp` extension remote setup to stop working as its image names were not changed and it depends on the same kyverno policies. This PR adapts the name of the images to be the same as the ones used in gardener. The changes here are similar to the ones in https://github.com/gardener/gardener-extension-registry-cache/pull/193

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/local-setup/kustomization.yaml
+++ b/local-setup/kustomization.yaml
@@ -1,7 +1,6 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 
-
 resources:
 - ../example
 
--- a/local-setup/kustomization.yaml
+++ b/local-setup/kustomization.yaml
@@ -1,0 +1,16 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+
+resources:
+- ../example
+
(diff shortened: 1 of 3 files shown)


local-setup/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+
```
ialidzhikov, line 4:
```suggestion
```

ialidzhikov (commented):
/lgtm

(the same changes as from https://github.com/gardener/gardener-extension-registry-cache/pull/193)

PR #138, opened 2024-07-12
Limit the memory of the rsyslog service

plkokanov:

/kind enhancement

This PR adds memory configurations for the `rsyslog.service` by adding a drop-in with the following content:

```
MemoryMin=15M
MemoryHigh=150M
MemoryMax=300M
MemorySwapMax=0
```

`MemoryMin` specifies the usual working memory of rsyslog, which is 15M - this memory should not be reclaimed by other services once rsyslog is restarted.
`MemoryHigh` specifies the memory after which the service will be throttled. 150M is the memory at which the rsyslog-relp action queue and the disk queue are full and the main queue of the rsyslog service is starting to get filled.
`MemoryMax` specifies the maximum amount of memory that the service can have after which it will be OOMKilled. The queues are usually full at around 200M-250M of memory, however this additional buffer is added in case of larger messages in the queues (queues get filled based on the number of messages, not their size).

The main reason for adding this PR, even though the queues that rsyslog uses are limited, is because there could still be memory leaks that could potentially lead to nodes getting all their memory consumed: https://github.com/rsyslog/rsyslog/issues/5318

Fixes #

**Release note**:

```other operator
The memory of the `rsyslog.service` systemd unit is now limited via a drop-in config. The following configurations are used: `MemoryMin=15M`, `MemoryHigh=150M`, `MemoryMax=300M`, `MemorySwapMax=0`
```

### DIFF
--- a/pkg/webhook/operatingsystemconfig/ensurer_test.go
+++ b/pkg/webhook/operatingsystemconfig/ensurer_test.go
@@ -346,6 +346,19 @@ func getRsyslogFiles(rsyslogConfig []byte, useExpectedContent bool) []extensions
 				},
 			},
 		},
+		{
+			Path:        "/etc/systemd/system/rsyslog.service.d/10-shoot-rsyslog-relp-memory-limits.conf",
+			Permissions: ptr.To(int32(0644)),
+			Content: extensionsv1alpha1.FileContent{
+				Inline: &extensionsv1alpha1.FileContentInline{
+					Data: getBasedOnCondition(useExpectedContent, `[Service]
+MemoryMin=15M
+MemoryHigh=150M
+MemoryMax=300M
+MemorySwapMax=0`, "old"),
+				},
+			},
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/webhook/operatingsystemconfig/ensurer_test.go
```diff
@@ -391,6 +391,22 @@ func getBasedOnCondition[T any](condition bool, whenTrue, whenFalse T) T {
 	return whenFalse
 }
 
+func getRsyslogDropIns(useExpectedContent bool) extensionsv1alpha1.Unit {
+	return extensionsv1alpha1.Unit{
+		Name: "rsyslog.service",
+		DropIns: []extensionsv1alpha1.DropIn{
```
ialidzhikov, line 397: TIL that there is such field in the OSC API. I was only aware of adding drop-ins in form of regular files: https://github.com/gardener/gardener/blob/6ba5765c27a1cb90e0fcb11764af7cfe6cc013cd/pkg/component/extensions/operatingsystemconfig/original/components/containerd/initializer.go#L91-L101.

ialidzhikov (commented):
/lgtm

pkg/webhook/operatingsystemconfig/ensurer_test.go
plkokanov: I decided to add the drop-in as a regular file as it feels safer. This way we avoid an unnecessary attempt by the `gardener-node-agent` to try to delete the `/etc/systemd/system/rsyslog.service` file (which in general does not exist, but might not be guaranteed for all cases) when the  extension is disabled.

ialidzhikov (commented):
/lgtm

PR #149, opened 2024-07-31
Expose audit rules configuration in the `shoot-rsyslog-relp` api 

plkokanov:

/kind enhancement

This PR makes it so that audit rules deployed by the extension are configurable. Additionally, it is now possible to disable configuring audit rules which means that the ones that are already available on the shoot nodes will be used.

The PR also adds a defaulting functionality for extension's provider config (`RsyslogRelpConfig`) so that deploying the audit rules is enabled by default. This is done to continue the current behaviour of the extension. 

Part of #2 

**Release note**:

```feature operator
A new api object `rsyslog-relp.extensions.gardener.cloud/v1alpha1.Auditd` is introduced which is used to specify a configuration for the linux audit daemon on the shoot nodes.
```
```feature operator
Two new fields have been added to the provider config for the `shoot-rsyslog-relp` extension:
- `.auditConfig.enabled` allows users to opt in whether to enable the reconfiguration of audit rules on the shoot's nodes and to also configure `auditd` to send logs to `rsyslog`. By default this field is `true`.
- `.auditConfig.configMapReferenceName` is a reference to a `ConfigMap` shoot resource which contains audit configuration. This field is only taken into account if `.auditRulesConfig.enabled` is `true`. The `ConfigMap` must contain a data key `auditd` which must contain a value of type `rsyslog-relp.extensions.gardener.cloud/v1alpha1.Auditd`.
```


pkg/controller/lifecycle/monitoring.go
```diff
@@ -55,45 +56,65 @@ func deployMonitoringConfig(ctx context.Context, c client.Client, namespace stri
...
+				},
+			},
+			{
+				Alert: "RsyslogRelpActionProcessingRateIsZero",
+				Expr:  intstr.FromString(`rate(rsyslog_pstat_processed{origin="core.action",name="rsyslog-relp"}[5m]) == 0`),
+				For:   ptr.To(monitoringv1.Duration("15m")),
+				Labels: map[string]string{
+					"service":    "rsyslog-relp",
+					"severity":   "warning",
+					"type":       "seed",
```
ialidzhikov, line 82:
```suggestion
					"type":       "shoot",
```

docs/usage/monitoring.md
```diff
@@ -119,9 +119,14 @@ Maximum size queue has reached.
 - Type: Gauge
 - Labels: `name` `node` `origin`
 
+#### rsyslog_augenrules_load_success
+Maximum size queue has reached.
```
ialidzhikov, line 123: This should be a copy-paste error.

pkg/controller/lifecycle/monitoring.go
```diff
@@ -55,45 +56,65 @@ func deployMonitoringConfig(ctx context.Context, c client.Client, namespace stri
...
+		if auditConfig != nil && ptr.Deref(auditConfig.Enabled, false) {
+			alertingRules = append(alertingRules, monitoringv1.Rule{
+				Alert: "RsyslogRelpAuditRulesNotLoadedSuccessfully",
+				Expr:  intstr.FromString(`absent(rsyslog_augenrules_load_success == 1)`),
+				For:   ptr.To(monitoringv1.Duration("15m")),
+				Labels: map[string]string{
+					"service":    "rsyslog-relp",
+					"severity":   "warning",
+					"type":       "seed",
+					"visibility": "operator",
```
ialidzhikov, line 101: I am not sure about the visibility. This should be rather `owner` or `all` as now Shoot owners can specify auditd rules via ConfigMap that gets referenced from the Shoot spec.

pkg/controller/lifecycle/monitoring.go
```diff
@@ -55,45 +56,65 @@ func deployMonitoringConfig(ctx context.Context, c client.Client, namespace stri
...
+		alertingRules := []monitoringv1.Rule{
+			{
+				Alert: "RsyslogTooManyRelpActionFailures",
+				Expr:  intstr.FromString(`sum(rate(rsyslog_pstat_failed{origin="core.action",name="rsyslg-relp"}[5m])) / sum(rate(rsyslog_pstat_processed{origin="core.action",name="rsyslog-relp"}[5m])) > bool 0.02 == 1`),
+				For:   ptr.To(monitoringv1.Duration("15m")),
+				Labels: map[string]string{
+					"service":    "rsyslog-relp",
+					"severity":   "warning",
+					"type":       "shoot",
+					"visibility": "operator",
```
ialidzhikov, line 68: Not introduced with this PR, but I am not sure that the visibility has to be `operator`. It should be rather `owner` as the target server is configured by the Shoot owner.

docs/usage/configuration.md
```diff
@@ -34,7 +34,7 @@ spec:
       apiVersion: rsyslog-relp.extensions.gardener.cloud/v1alpha1
       kind: RsyslogRelpConfig
       # Set the target server to which logs are sent. The server must support the RELP protocol.
-      target: some.rsyslog-rlep.server
+      target: some.rsyslog-relp.server
```
ialidzhikov, line 37: You could also fix similar typo in https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/c8be0cbeba2f36ead53222546f0a9b2e7edde1d8/pkg/cmd/rsyslogrelp/options.go#L5.

docs/usage/configuration.md
```diff
@@ -184,4 +184,77 @@ spec:
...
+auditRules: |
+  ## First rule - delete all existing rules
+  -D
+  ## Now define some custom rules
+  -a exit,always -F arch=b64 -S setuid -S setreuid -S setgid -S setregid -F auid>0 -F auid!=-1 -F key=privilege_escalation
+  -a exit,always -F arch=b64 -S execve -S execveat -F euid=0 -F auid>0 -F auid!=-1 -F key=privilege_escalation
+```
+
+To deploy this configuration, it must be embedded in a config map. **Note** that the data key storing this configuration must be named `auditd`.
+An example `ConfigMap` is given below:
```
ialidzhikov, line 208:
```suggestion
To deploy this configuration, it must be embedded in an immutable ConfigMap.

> [!NOTE]
> The data key storing this configuration must be named `auditd`.

An example `ConfigMap` is given below:
```

docs/usage/configuration.md
```diff
@@ -184,4 +184,77 @@ spec:
...
+        configMapReferenceName: audit-config
+  resources:
+    - name: audit-config
+      resourceRef:
+        apiVersion: v1
+        kind: ConfigMap
+        name: audit-config-v1
+```
+
+Finally, by setting `providerConfig.auditConfig.enabled` to `false` in the `shoot-rsyslog-relp` extension configuration, you can completely disable the additional audit configurations performed by the extension.
```
ialidzhikov, line 260: IIUC, there are audit rules coming from 3 sources: rules that come with the host OS, rules added by default by the extension and rules that can now be added by the Shoot owners.

When `providerConfig.auditConfig.enabled` won't deploy any rules (extension default ones and Shoot owners ones), right? We can be more explicit here as 
> [...] you can completely disable the additional audit configurations performed by the extension.

it is not super clear what you disable.

docs/usage/configuration.md
```diff
@@ -184,4 +184,77 @@ spec:
 You can set a few additional parameters for the TLS connection: `.tls.authMode`, `tls.permittedPeer`, and `tls.tlsLib`. Refer to the rsyslog documentation for more information on these parameters:
 - [`.tls.authMode`](https://www.rsyslog.com/doc/v8-stable/configuration/modules/omrelp.html#tls-authmode)
 - [`.tls.permittedPeer`](https://www.rsyslog.com/doc/v8-stable/configuration/modules/omrelp.html#tls-permittedpeer)
-- [`.tls.tlsLib`](https://www.rsyslog.com/doc/v8-stable/configuration/modules/imrelp.html#tls-tlslib)
\ No newline at end of file
+- [`.tls.tlsLib`](https://www.rsyslog.com/doc/v8-stable/configuration/modules/imrelp.html#tls-tlslib)
+
+### Configuring the Audit Daemon on the Shoot Nodes
```
ialidzhikov, line 189: Does it make sense to also add the new fields in the Shoot spec in https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/b6e6bb8a1091c3443d36efbfda3fb60bd18b23b8/docs/usage/configuration.md#shoot-configuration?

example/configmap-rsyslog-auditd.yaml
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+---
+kind: ConfigMap
```
ialidzhikov, line 6: We could also adapt the example Shoot spec to contain reference to this ConfigMap. The same way as for the Secret: https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/71f15814a648731897101c2667b352efbcaca27b/example/shoot.yaml#L41-L46

For testing it would be much easier to uncomment instead of building the needed partial spec again.

test/common/testdata/audit/audit-configmap.yaml
```diff
@@ -0,0 +1,25 @@
+# SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
```
ialidzhikov, line 3: I would rather add https://github.com/gardener/gardener-extension-registry-cache/blob/757bb0231c3b5539eb766b71871717b97ecd532a/.reuse/dep5#L31 to the `.reuse/dep5` file.

test/common/verifier.go
```diff
@@ -23,6 +23,29 @@ import (
 
 const fileForAuditEvent = "/etc/newfile"
 
+// ExpectedAuditRules are the expected audit rules. Exposed so that they can be modified by tests.
+var ExpectedAuditRules = `## This file is automatically generated from /etc/audit/rules.d
```
ialidzhikov, line 27: Wouldn't it be better if expected audit rules are somehow configured in the verifier (passed outside, from the caller) instead of overwriting an exported var?

pkg/apis/rsyslog/types_rsyslog.go
```diff
@@ -62,6 +65,16 @@ type LoggingRule struct {
 	Severity int
 }
 
+// AuditConfig contains options to configure the audit system.
+type AuditConfig struct {
+	// Enabled determines whether auditing configurations are applied to the nodes or not.
+	// Defaults to true.
+	Enabled *bool
```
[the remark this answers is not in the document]
plkokanov: Yes, I think that makes sense. For some reason I thought that I wouldn't be able to default it to true, but you're right that that isn't a problem since `AuditConfig` is optional (can be nil).

pkg/apis/rsyslog/v1alpha1/types_auditd.go
```diff
@@ -0,0 +1,20 @@
...
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Auditd contains configuration for the audit daemon.
+type Auditd struct {
+	metav1.TypeMeta
+
+	// AuditRules contains the audit rules that will be placed under /etc/audit/rules.d.
+	AuditRules string `json:"auditRules,omitempty"`
```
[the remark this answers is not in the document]
plkokanov: I added the validation to check that `AuditRules` are not empty later on and forgot to reflect it here. Thanks!

pkg/apis/rsyslog/types_rsyslog.go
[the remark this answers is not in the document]
plkokanov: Actually, it might be better flip it and rename the field to `Disabled` - this way users have to explicitly specify that they do not want audit rules and hence skip on this security feature. @ialidzhikov wdyt?

pkg/webhook/operatingsystemconfig/auditd.go
```diff
@@ -74,3 +131,16 @@ func getAuditdFiles() []extensionsv1alpha1.File {
...
+	auditConfig, _, err := decoder.Decode([]byte(data), nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	switch config := auditConfig.(type) {
+	case *rsyslog.Auditd:
+		return []byte(config.AuditRules), nil
+	}
+	return nil, nil
```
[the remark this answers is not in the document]
plkokanov: I changed this function a couple of times and initially, I was planning to have other entries, appart from `auditd` config to this configmap - e.g. `iptables` so this is sort of a leftover. I'll optimise it now as you suggested.

pkg/controller/lifecycle/monitoring.go
```diff
@@ -55,45 +56,65 @@ func deployMonitoringConfig(ctx context.Context, c client.Client, namespace stri
...
+				For:   ptr.To(monitoringv1.Duration("15m")),
+				Labels: map[string]string{
+					"service":    "rsyslog-relp",
+					"severity":   "warning",
+					"type":       "seed",
+					"visibility": "operator",
+				},
+				Annotations: map[string]string{
+					"description": "The rsyslog augenrules load success is 0 meaning that there was an error when calling 'augenrules --load' on the Shoot nodes",
+					"summary":     "Rsyslog augenrules load success is 0",
```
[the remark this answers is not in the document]
plkokanov: 0 should be the correct string here - it indicates that `augenrules --load` was not called successfully. Note that the expression above contains the `absent` keyword:
```
absent(rsyslog_augenrules_load_success == 1)
```

Meaning that if we do not find this metric with value 1 we will fire an alert.

pkg/controller/lifecycle/monitoring.go
plkokanov: I think `all` makes sense as we might also be interested in seeing these alerts if shoot owners miss them.

test/common/testdata/audit/audit-configmap.yaml
plkokanov: Then I would do this for the entire examples folder. Other files here also contain the licensing header.

test/common/verifier.go
plkokanov: Yeah, I was considering that, since they only differ in one test case I left a "default" variable here so that I don't have to pass them everywhere. But I could change that to keep it in line with other verifier options.

pkg/apis/rsyslog/types_rsyslog.go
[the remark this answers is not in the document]
ialidzhikov: As we talked, let's keep it enable, if possible. `disabled: false` is a double negation which one idea harder to follow.

docs/usage/configuration.md
ialidzhikov: Just to reply to myself:
- The following deploys the extension default audit rules as of today
```yaml
      auditConfig:
        enabled: true
```

- The following deploys only the rules specified in the ConfigMap
```yaml
      auditConfig:
        enabled: true
        configMapReferenceName: audit-config
```

- The following does not deploy any audit rules
```yaml
      auditConfig:
        enabled: false
        configMapReferenceName: audit-config
```

pkg/controller/lifecycle/monitoring.go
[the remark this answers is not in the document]
ialidzhikov: You are right, my bad 🤦

docs/usage/configuration.md
plkokanov: Good point, I could actually list the options like that. I tried simply modifying the text but this is clearer, thanks!

test/common/verifier.go
```diff
@@ -23,8 +23,10 @@ import (
 
 const fileForAuditEvent = "/etc/newfile"
 
-// ExpectedAuditRules are the expected audit rules. Exposed so that they can be modified by tests.
-var ExpectedAuditRules = `## This file is automatically generated from /etc/audit/rules.d
+// ExpectedAuditRules is a set of expected audit rules to be deployed on the shoot nodes
+// when the AuditConfig in the shoot-rsyslog-relp extension providerConfig is enabled and
+// does not specify a ConfigMap reference containing custom audit rules.
+const ExpectedAuditRules = `## This file is automatically generated from /etc/audit/rules.d
```
ialidzhikov, line 29: Nit: I would move the const to a private const in `test/testmachinery/shoot/enable_disable_test.go`. Right now, it is used only there. 

I would image something like:

```golang
const (
	defaultExtensionAuditRules = "..."
	customAuditRules = "..."
)
```

ialidzhikov (commented):
Final suggestion, rest lgtm

Thanks for addressing all the feedback!

ialidzhikov (approved):
/lgtm

ialidzhikov (approved):
/lgtm

PR #150, opened 2024-07-31
Adds missing empty new line at end of audit rules file

plkokanov:

/kind bug

This PR adds a new line at the  end of https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/main/pkg/webhook/operatingsystemconfig/testdata/11-privileged-special.rules
Generally, separate rule files should always end with a new line, otherwise they are not merged together properly by `augenrules --load`.

This will be properly tested by a tm test added in #149 

Fixes #

**Release note**:

```bugfix operator
Fixed an issue that caused the `-a exit,always -F arch=b64 -S mount_setattr -F auid!=-1 -F key=privileged_special` audit rule to not get correctly applied.
```


ialidzhikov (approved):
/lgtm

Issue #153, opened 2024-08-07
[Flaky Test] TestMachinery tests can fail with `failed to resolve shoot resource reference "rsyslog-relp-tls": secrets "rsyslog-relp-tls" not found`


plkokanov:
**How to categorize this issue?**

/area testing
/kind flake

**Which test(s)/suite(s) are flaking**:
The testmachinery tests (under [this](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/tree/main/test/testmachinery/shoot) directory) are currently flaky with the following error:

```
2024-08-07 03:19:19	
2024-08-07T00:19:19.351321758Z stdout F   In [It] at: /src/test/testmachinery/shoot/enable_disable_test.go:39 @ 08/07/24 00:19:15.52
2024-08-07 03:19:19	
2024-08-07T00:19:19.351317871Z stdout F       }
2024-08-07 03:19:19	
2024-08-07T00:19:19.351313955Z stdout F           },
2024-08-07 03:19:19	
2024-08-07T00:19:19.351309935Z stdout F               },
2024-08-07 03:19:19	
2024-08-07T00:19:19.351305204Z stdout F                   Code: 403,
2024-08-07 03:19:19	
2024-08-07T00:19:19.351301253Z stdout F                   },
2024-08-07 03:19:19	
2024-08-07T00:19:19.351297654Z stdout F                       RetryAfterSeconds: 0,
2024-08-07 03:19:19	
2024-08-07T00:19:19.351293925Z stdout F                       Causes: nil,
2024-08-07 03:19:19	
2024-08-07T00:19:19.351290362Z stdout F                       UID: "",
2024-08-07 03:19:19	
2024-08-07T00:19:19.351286575Z stdout F                       Kind: "shoots",
2024-08-07 03:19:19	
2024-08-07T00:19:19.35128282Z stdout F                       Group: "core.gardener.cloud",
2024-08-07 03:19:19	
2024-08-07T00:19:19.351278506Z stdout F                       Name: "tmive-twa",
2024-08-07 03:19:19	
2024-08-07T00:19:19.351271782Z stdout F                   Details: {
2024-08-07 03:19:19	
2024-08-07T00:19:19.351129943Z stdout F                   Reason: "Forbidden",
2024-08-07 03:19:19	
[...]

PR #164, opened 2024-08-28
Testmachinery test uses rsyslog-tls secret with unique name

Kostov6:

/area robustness
/kind flake

Testmachinery test now generates rsyslog-relp-tls secret to have a unique suffix that will prevent collisions with other tests

Fixes https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/issues/153

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/testmachinery/shoot/enable_disable_test.go
+++ b/test/testmachinery/shoot/enable_disable_test.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
-	testutils "github.com/gardener/gardener/pkg/utils/test"
 	. "github.com/gardener/gardener/pkg/utils/test/matchers"
 	"github.com/gardener/gardener/test/framework"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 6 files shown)


test/testmachinery/shoot/enable_disable_test.go
```diff
@@ -145,9 +145,12 @@ var _ = Describe("Shoot rsyslog-relp testing", func() {
 			defer cancel()
 
 			var err error
-			createdResources, err = testutils.EnsureTestResources(ctx, f.GardenClient.Client(), f.ProjectNamespace, "../../common/testdata/tls")
+			createdResources, err = testutils.ReadTestResources(f.GardenClient.Client().Scheme(), f.ProjectNamespace, "../../common/testdata/tls")
```
plkokanov, line 148:
```suggestion
			resourcesToCreate, err = testutils.ReadTestResources(f.GardenClient.Client().Scheme(), f.ProjectNamespace, "../../common/testdata/tls")
```

test/testmachinery/shoot/enable_disable_test.go
```diff
@@ -145,9 +145,12 @@ var _ = Describe("Shoot rsyslog-relp testing", func() {
...
 			Expect(err).NotTo(HaveOccurred())
 			Expect(len(createdResources)).ToNot(BeZero())
+			for _, obj := range createdResources {
```
plkokanov, line 151:
```suggestion
			for _, obj := range resourcesToCreate {
```

plkokanov (changes requested):
You also have to make a similar change for the audit configmap https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/1a8c2d32c67b62940980b9bd38660c5f07bd1dfa/test/common/testdata/audit/audit-configmap.yaml#L4 to use `generateName`
Also adapt where it is used: https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/1a8c2d32c67b62940980b9bd38660c5f07bd1dfa/test/testmachinery/shoot/enable_disable_test.go#L190

The secret resource is also used here: https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/1a8c2d32c67b62940980b9bd38660c5f07bd1dfa/test/e2e/create_enable_disable_delete.go#L112
so this has to be adapted as well. 

Let's expose the test data directories (`common/testdata/tls` and `common/testdata/audit/`) as constants in the `test/common` package so that it's easier to see all places where they are used

Also can we combine the calls to `ReadTestResources` and the calls to `Create` in a common function?

PR #171, opened 2024-09-13
Handle `-- Mark --` logs in `rsyslog-relp-echo-server`'s output when checking if the extension is disabled

plkokanov:

/kind flake

This PR makes it so that `-- Mark --` messages, which are output by the `rsyslog-relp-echo-server` to indicate rsyslog is working properly, are taken into account when checking if no logs are sent from the nodes to the echo server: [ref](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/1edcd373b1e7614a68d0fe4d3dbb1952c31b4d58/test/common/verifier.go#L169-L174).

Fixes #

**Release note**:

```other operator
NONE
```

### DIFF
--- a/test/common/verifier.go
+++ b/test/common/verifier.go
@@ -170,7 +170,7 @@ func (v *Verifier) verifyThatLogsAreNotForwardedToEchoServer(ctx context.Context
 	EventuallyWithOffset(2, func(g Gomega) {
 		logLines, err := v.getLogs(ctx, timeBeforeLogGeneration)
 		g.Expect(err).NotTo(HaveOccurred())
-		g.Expect(logLines).To(BeEmpty())
+		g.Expect(logLines).To(Or(BeEmpty(), ConsistOf(ContainSubstring("-- Mark --"))))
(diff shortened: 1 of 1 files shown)


test/common/verifier.go
```diff
@@ -170,7 +170,7 @@ func (v *Verifier) verifyThatLogsAreNotForwardedToEchoServer(ctx context.Context
 	EventuallyWithOffset(2, func(g Gomega) {
 		logLines, err := v.getLogs(ctx, timeBeforeLogGeneration)
 		g.Expect(err).NotTo(HaveOccurred())
-		g.Expect(logLines).To(BeEmpty())
+		g.Expect(logLines).To(Or(BeEmpty(), ConsistOf(ContainSubstring("-- Mark --"))))
```
ialidzhikov, line 173: Optionally, we can optionally add a comment describing the `-- Mark --` logs as I didn't know what is the reason behind it:

```suggestion
		// -- Mark -- is a period log emitted by rsyslogd denoting that there are no incoming logs.
		g.Expect(logLines).To(Or(BeEmpty(), ConsistOf(ContainSubstring("-- Mark --"))))
```

plkokanov: The more correct description would be that it is a "heartbeat" log message which lets you know that it is working. I'll add it for clarity.

ialidzhikov (approved):
/lgtm

PR #189, opened 2024-10-29
Enable gosec for static application security testing

Kostov6:

/area compliance
/kind enhancement

This PR enables `gosec` following https://github.com/gardener/gardener-extension-shoot-lakom-service/pull/116/commits/03080233488e028663ae36ff1be631f241585579

**Release note**:

```noteworthy developer
`gosec` is made available for SAST(static application security testing), it can be run with `make sast` or `make sast-report`, but is also incorporated in the `verify` and `verify-extended` makefile targets. 
```

### DIFF
--- a/hack/sast.sh
+++ b/hack/sast.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+#
+# SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+root_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"
+
+gosec_report="false"
+gosec_report_parse_flags=""
+
+parse_flags() {
+  while test $# -gt 1; do
+    case "$1" in
+      --gosec-report)
+        shift; gosec_report="$1"
+        ;;
+      *)
+        echo "Unknown argument: $1"
+        exit 1
+        ;;
+    esac
+    shift
+  done
+}
+
+parse_flags "$@"
+
+echo "> Running gosec"
+gosec --version
+if [[ "$gosec_report" != "false" ]]; then
+  echo "Exporting report to $root_dir/gosec-report.sarif"
+  gosec_report_parse_flags="-track-suppressions -fmt=sarif -out=gosec-report.sarif -stdout"
+fi
+
+# Gardener uses code-generators https://github.com/kubernetes/code-generator and https://github.com/protocolbuffers/protobuf
+# which create lots of G103 (CWE-242: Use of unsafe calls should be audited) & G104 (CWE-703: Errors unhandled) errors.
+# However, those generators are best-pratice in Kubernetes environment and their results are tested well.
+# Thus, generated code is excluded from gosec scan.
+# Nested go modules are not supported by gosec (see https://github.com/securego/gosec/issues/501), so the ./hack folder
+# is excluded too. It does not contain productive code anyway.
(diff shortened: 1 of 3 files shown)


hack/sast.sh
```diff
@@ -0,0 +1,44 @@
...
+  gosec_report_parse_flags="-track-suppressions -fmt=sarif -out=gosec-report.sarif -stdout"
+fi
+
+# Gardener uses code-generators https://github.com/kubernetes/code-generator and https://github.com/protocolbuffers/protobuf
+# which create lots of G103 (CWE-242: Use of unsafe calls should be audited) & G104 (CWE-703: Errors unhandled) errors.
+# However, those generators are best-pratice in Kubernetes environment and their results are tested well.
+# Thus, generated code is excluded from gosec scan.
+# Nested go modules are not supported by gosec (see https://github.com/securego/gosec/issues/501), so the ./hack folder
+# is excluded too. It does not contain productive code anyway.
+gosec -exclude-generated -exclude-dir=hack $gosec_report_parse_flags ./...
```
plkokanov, line 44: Developers could use the `ci-e2e-kind.sh` script to run all e2e tests locally. As part of that `gardener/gardener` is downloaded under the `/gardener` directory. refs:
https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/065773e4ab8ac9e2dc2156a2eed2a88fc5c3f613/hack/ci-e2e-kind.sh#L15-L19
https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/065773e4ab8ac9e2dc2156a2eed2a88fc5c3f613/.gitignore#L17

This could lead to errors when running gosec locally due to nested modules. Could we also exclude that directory?

hack/sast.sh
```diff
@@ -0,0 +1,44 @@
...
+parse_flags "$@"
+
+echo "> Running gosec"
+gosec --version
+if [[ "$gosec_report" != "false" ]]; then
+  echo "Exporting report to $root_dir/gosec-report.sarif"
+  gosec_report_parse_flags="-track-suppressions -fmt=sarif -out=gosec-report.sarif -stdout"
+fi
+
+# Gardener uses code-generators https://github.com/kubernetes/code-generator and https://github.com/protocolbuffers/protobuf
```
plkokanov, line 38:
```suggestion
# The shoot-rsyslog-relp extension uses code-generators https://github.com/kubernetes/code-generator and https://github.com/protocolbuffers/protobuf
```

plkokanov (changes requested):

hack/sast.sh
ialidzhikov: Nit: We don't have protobuf in extensions.

ialidzhikov: But in general instead of copying this script file on `make tidy` we could call it directly from the hack dir of the Gardener the same way we call other scripts.

plkokanov: I'm ok to not copy it locally under hacks. I think it was done for other scripts so that they can be conveniently called from the hack directory as there is no make target for them. 
Btw, It might be useful to extend the original script under g/g so that additional parameters can be passed to it - e.g. projects that depend on it might have additional directories that they want to exclude.

ialidzhikov: > Btw, It might be useful to extend the original script under g/g so that additional parameters can be passed to it - e.g. projects that depend on it might have additional directories that they want to exclude.

Yep, we also want to exclude the gardener/ dir in the registry-cache repo. Ref https://github.com/gardener/gardener-extension-registry-cache/pull/272#discussion_r1824144071

plkokanov:
Generally /lgtm but we need to merge https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/pull/196 first

PR #243, opened 2025-03-04
Enable filtering on log message content

### LINKED ISSUE #144: Enhance the logging rules so that filtering can be performed on log message content
**How to categorize this issue?**

/kind enhancement

Add a field in the `rsyslogRelpConfig.loggingRules` that allows filtering based on message content. The filters should allow for both excluding or including messages based on their content via regular expressions. This could be done by adding `regex` and `exclude` fields which should be pointers to strings and marked as optional. Below is an example of what the new api would look like:

```
apiVersion: rsyslog-relp.extensions.gardener.cloud/v1alpha1
kind: RsyslogRelpConfig
target: localhost
port: 1520
loggingRules:
- severity: 6
  programNames: ["kubelet"]
  messageContent:
    regex: "some message content"
    exclude: "some other message content"
``` 

To implement the regular expression, the following [`RainerScript`](https://www.rsyslog.com/doc/rainerscript/index.html) functions could be used in the rsyslog config: [`re_match_i()`](https://www.rsyslog.com/doc/rainerscript/functions/rs-re_match_i.html) - case insensitive match or [`re_match()`](https://www.rsyslog.com/doc/rainerscript/functions/rs-re_match.html) - case sensitive match.

This would result in the following rsyslog config:
```
if $programname == ["kubelet"] and $syslogseverity <= 6 and re_match($msg,'some message content') == 1 and re_match(&msg, 'some other message content') == 0 then {
  call relp_action_ruleset
  stop
}
```

This means that `messageContent` `include` and `exclude` directives are `and`-ed when they are part of the same `loggingRule`. A subsequent `loggingRule` could contain the same `programNames` and same `severity` but different `messageContent` fields resulting in an `or` operation - this is because `loggingRules[]` are checked in-order.

Currently the `loggingRules[].severity` field is always required. However, we could change that and make at least one of `severity`, `programNames` or `messageContent` be required.

**Why is this needed**:
Currently, the `shoot-rsyslog-relp` extension only allows for logs to be filtered based on the program name (that generated them) and their severity. However, this is not enough as sometimes users might want to further fine tune the filtering based on the message content of the logs.

RadaBDimitrova:

/area dev-productivity
/kind enhancement

Adds logic for filtering the logs by their message content.

Fixes #144

N/A

**Release note**:

```other operator
Introduced field `messageContent` defines regular expressions for including and excluding logs based on their message content.
```


Kostov6 (changes requested):
Please add test case to `ensurer_test.go`

Kostov6 (commented):
```
// MessageContent contains fields for including and excluding messages
type MessageContent struct {
	// MatchRegex is a regex used when deciding what messages to include
	IncludeRegex *string
	// ExcludeRegex is a regex used when deciding what messages to exclude
	ExcludeRegex *string
}
```
@plkokanov isn't this more clearer? Because on first read I thought that `Exclude` did string matching

plkokanov:
> ```
> // MessageContent contains fields for including and excluding messages
> type MessageContent struct {
> 	// IncludeRegex is a regex used when deciding what messages to include
> 	IncludeRegex *string
> 	// ExcludeRegex is a regex used when deciding what messages to exclude
> 	ExcludeRegex *string
> }
> ```
> 
> @plkokanov isn't this more clearer? Because on first read I thought that `Exclude` did string matching

I "stole" the fields' names from the fluentbit grep filter config: https://docs.fluentbit.io/manual/pipeline/filters/grep 
Tbh I favour `regex` and `exclude` a bit more to keep them short.

pkg/apis/rsyslog/validation/validation_test.go
```diff
@@ -111,6 +111,26 @@ var _ = Describe("Validation", func() {
 			Expect(errorList).To(matcher)
 		})
 
+		It("should not allow empty no field of a logging rule to be set", func() {
+			config := rsyslog.RsyslogRelpConfig{
+				Target:       relpTarget,
+				Port:         relpTargetPort,
+				LoggingRules: []rsyslog.LoggingRule{{}},
```
Kostov6, line 118: Can you add a valid logging rule and a empty one for this scenario?

Kostov6 (commented):
Can you also add documentation for this?

Kostov6 (commented):
/lgtm

ialidzhikov:
Could you rebase after https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/pull/248 was merged?
This should hopefully fix the e2e test.

plkokanov (changes requested):
Please also add validation for the regular expressions in https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/96dd6874e53eeb93de6ca37fb6a04f764968b2f8/pkg/apis/rsyslog/validation/rsyslog_relp.go#L90
You can use `regexp.CompilePOSIX` and check if it returns an error for the validation. This checks for `POSIX ERE (egrep)` syntax which is also the one supported by rsyslog's `re_match` function.

Additionally, add validation that at least one of `messageContent.regex` or `messageContent.exclude` has to be specified.

Add tests for the new validations.

pkg/apis/rsyslog/validation/rsyslog_relp.go
```diff
@@ -86,6 +89,27 @@ func validateLoggingRules(loggingRules []rsyslog.LoggingRule, fldPath *field.Pat
...
+		for index, rule := range loggingRules {
+			if len(rule.ProgramNames) == 0 && rule.Severity == nil && rule.MessageContent == nil {
+				allErrs = append(allErrs, field.Required(fldPath.Index(index), "at least one of .programNames, .messageContent, or .severity is required"))
+			}
+			if rule.MessageContent != nil {
+				if rule.MessageContent.Regex == nil && rule.MessageContent.Exclude == nil {
+					allErrs = append(allErrs, field.Required(fldPath.Index(index).Child("messageContent"), "either .regex or .exclude has to be provided"))
+				}
+				if regex := rule.MessageContent.Regex; regex != nil {
+					if _, err := regexp.CompilePOSIX(*regex); err != nil {
```
plkokanov, line 102: Let's use `strconv.Quote` to process the regular expression before compiling it:
```suggestion
					quotedRegex := strconv.Quote(*regex)
					if _, err := regexp.CompilePOSIX(quotedRegex); err != nil {
```

P.S. You could probably extract the check in a method so that it can be reused when checking both `.regex` and `.exclude`

pkg/apis/rsyslog/validation/rsyslog_relp.go
```diff
@@ -86,6 +89,27 @@ func validateLoggingRules(loggingRules []rsyslog.LoggingRule, fldPath *field.Pat
...
+						allErrs = append(allErrs, field.Required(fldPath.Index(index).Child("messageContent").Child("regex"), fmt.Sprintf("not a valid POSIX ERE regular expression: %v", err)))
+					}
+				}
+				if exclude := rule.MessageContent.Exclude; exclude != nil {
+					if _, err := regexp.CompilePOSIX(*exclude); err != nil {
```
plkokanov, line 107: Same as above

pkg/webhook/operatingsystemconfig/rsyslog.go
```diff
@@ -280,15 +280,25 @@ func computeLogFilters(loggingRules []rsyslog.LoggingRule) []string {
...
-		} else {
-			filters = append(filters, fmt.Sprintf("$syslogseverity <= %d", rule.Severity))
+			currentFilters = append(currentFilters, fmt.Sprintf("$programname == [%s]", strings.Join(programNames, ",")))
+		}
+		if rule.Severity != nil {
+			currentFilters = append(currentFilters, fmt.Sprintf("$syslogseverity <= %d", *rule.Severity))
 		}
+		if rule.MessageContent != nil {
+			if include := rule.MessageContent.Regex; include != nil {
+				currentFilters = append(currentFilters, fmt.Sprintf("re_match($msg,'%s') == 1", *include))
```
plkokanov, line 295:
```suggestion
				quotedRegex := strconv.Quote(*include)
				currentFilters = append(currentFilters, fmt.Sprintf("re_match($msg, %s) == 1", quotedRegex))
```

pkg/webhook/operatingsystemconfig/rsyslog.go
```diff
@@ -280,15 +280,25 @@ func computeLogFilters(loggingRules []rsyslog.LoggingRule) []string {
...
+			}
+			if exclude := rule.MessageContent.Exclude; exclude != nil {
+				currentFilters = append(currentFilters, fmt.Sprintf("re_match($msg,'%s') == 0", *exclude))
```
plkokanov, line 298: Same as above.

plkokanov (commented):
Thanks a lot for working on this topic and addressing my comments :) 

/lgtm
/approve

PR #251, opened 2025-03-21
Bring back tm beta labels so that tests can be properly GAed

plkokanov:

/kind enhancement

This PR brings back the `beta` and `shoot` labels to the TM test definition. This is currently necessary because our test landscapes use the `IstioTLSTermination` feature gate which requires changes in the `RootPodExecutor` interface used by the `shoot-rsyslog-relp` tests. Those changes come in g/g `v1.114`. 
Without these labels, when the tests are executed on the landscapes, they are picked from the `v0.7.0` release of the `shoot-rsyslog-relp` extension which does not have the required changes to work with `IstioTLSTermination`. Adding these labels will let TM find the test definitions from the `main` branch of this repository and use them.

Fixes #

**Release note**:

```other operator
NONE
```


Kostov6 (commented):
/lgtm

PR #301, opened 2025-08-28
Add generation for `Extension` example

timuthy:

/area ops-productivity
/kind enhancement

This PR adds the generation for an `Extension` example manifest, similar to [`example/controller-registration.yaml`](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/master/example/controller-registration.yaml) which will be replaced on the long run.

**Release note**:

```other operator
An example `Extension` manifest for extension registration has been added. It can be found at [`example/extension.yaml`](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/v0.11.0/example/extension.yaml).
```


plkokanov (changes requested):
Thanks for the PR!

Could you apply the following change to `REUSE.toml`:

```patch
diff --git a/REUSE.toml b/REUSE.toml
index afeca8b..9b131b6 100644
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -20,10 +20,10 @@ path = [
     "charts/gardener-extension-shoot-rsyslog-relp-admission/charts/application/templates/_helpers.tpl",
     "charts/gardener-extension-shoot-rsyslog-relp-admission/charts/runtime/.helmignore",
     "charts/gardener-extension-shoot-rsyslog-relp-admission/charts/runtime/templates/_helpers.tpl",
-    "example/extension/.gitignore",
-    "example/extension/extension-config-patch.yaml.tmpl",
+    "example/extension/**",
     "example/local/charts/rsyslog-relp-echo-server/.helmignore",
     "example/controller-registration.yaml",
+    "example/extension.yaml",
     "example/kustomization.yaml",
     "hack/api-reference/**.json",
     "hack/cherry-pick-pull.sh",
```

timuthy:
Thanks for the review @plkokanov. I've addressed your feedback, PTAL.

Makefile
```diff
@@ -87,6 +87,7 @@ tidy:
 .PHONY: clean
 clean:
 	@$(shell find ./example -type f -name "controller-registration.yaml" -exec rm '{}' \;)
+	@$(shell find ./example -type f -name "extension.yaml" -exec rm '{}' \;)
```
ialidzhikov, line 90: Don't we need to rm `example/extension/base/extension.yaml` as well? 😅 

At least, we did it for the registry-cache: ref https://github.com/gardener/gardener-extension-registry-cache/pull/448/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R66

timuthy: The command here already deletes both files. I missed to adjust the registry cache extension after I force pushed, so it's currently redundant there.

ialidzhikov: I see, thanks for explaining! I will raise a PR for the registry-cache extension to unify. 👍

ialidzhikov: Ref https://github.com/gardener/gardener-extension-registry-cache/pull/451

plkokanov (commented):
Thanks again @timuthy for contributing this and @ialidzhikov with reviewing!

/approve

PR #303, opened 2025-09-01
Fix orphaned resources after testmachinery tests fail

plkokanov:

/area testing
/kind bug

This PR fixes an issue that could cause the `rsyslog-relp-tls` `Secret` or the `audit-config` `ConfigMap` that are created during testmachinery tests to not get deleted if the testmachinery tests fail.

This issue could occur in the following cases:
- If the test cleanup context times out while waiting for the `Shoot` to be reconciled after the `shoot.spec.extensions` and the `shoot.spec.resources` have been modified [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/b7240360185ed33457e6db7c1f3772aa3e8878ef/test/testmachinery/shoot/enable_disable_test.go#L166-L171) and [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/b7240360185ed33457e6db7c1f3772aa3e8878ef/test/testmachinery/shoot/enable_disable_test.go#L207-L212)
- If the update of the `Shoot` fails after the modification of `shoot.spec.extensions` and `shoot.spec.resources` [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/b7240360185ed33457e6db7c1f3772aa3e8878ef/test/testmachinery/shoot/enable_disable_test.go#L166-L171) and [here](https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/b7240360185ed33457e6db7c1f3772aa3e8878ef/test/testmachinery/shoot/enable_disable_test.go#L207-L212), the code that cleans up the `Secret` and `ConfigMap` is never called

To fix this
- I have increased the test cleanup timeout from 10 to 12 minutes
- The shoot spec update now uses a child context which times out after 10 minutes. This leaves 2 minutes for the `Secret` and `ConfigMap` to be deleted.
- The `Secret` and `ConfigMap` are now deleted in a `defer` call so that the cleanup is called even if the shoot update does not succeed. Note that the resources might not get cleaned immediately if they were not removed from the `shoot.spec.resources` array. However, the shoot that was used for the testmachinery test will eventually be deleted and the resources should then also get successfully deleted afterwards.

Fixes #

I have used `defer` because `DeferCleanup` does not seem to work with `framework.WithCAfterTest` 

**Release note**:

```other operator
NONE
```


RadaBDimitrova (commented):
/lgtm

PR #308, opened 2025-09-12
Enhance validation of `RsyslogRelpConfig`

RadaBDimitrova:

/area quality
/kind enhancement

This PR introduces additional validation for the `RsyslogRelpConfig` by validating the `target` and `permittedPeer` values using `RFC 1123` specifications. It also includes the extra validation for `programNames` which should not contain  `[`, `:` or `/`.

N/A

Here we need to fine-tune our limitations for the `target`
https://github.com/gardener/gardener-extension-shoot-rsyslog-relp/blob/72b832a902fb6bbb527c7d1d9fe322e2725422b0/pkg/apis/rsyslog/validation/rsyslog_relp.go#L36-L38
since there aren't many limitations listed in the [rsyslog documentation](https://www.rsyslog.com/doc/configuration/modules/omrelp.html#target).
/cc @plkokanov @Kostov6 

**Release note**:

```other operator
NONE
```

### DIFF
--- a/pkg/apis/rsyslog/validation/rsyslog_relp.go
+++ b/pkg/apis/rsyslog/validation/rsyslog_relp.go
@@ -142,6 +142,9 @@ func validateLoggingRules(loggingRules []rsyslog.LoggingRule, fldPath *field.Pat
 func validateProgramNames(programNames []string, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	for index, name := range programNames {
+		if name == "" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), name, ".programNames can't contain empty program name"))
+		}
 		if invalidCharactersForProgramNameRegex.MatchString(name) {
 			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), name, ".programNames can't contain `[`, `:` or `/`"))
 		}
--- a/pkg/apis/rsyslog/validation/validation_test.go
+++ b/pkg/apis/rsyslog/validation/validation_test.go
@@ -163,6 +163,26 @@ var _ = Describe("Validation", func() {
 			Expect(errorList).To(matcher)
 		})
 
+		It("should not allow a logging rule with set programNames to have empty names", func() {
+			config := rsyslog.RsyslogRelpConfig{
+				Target:       relpTarget,
+				Port:         relpTargetPort,
+				LoggingRules: []rsyslog.LoggingRule{{ProgramNames: []string{""}, Severity: ptr.To(4)}},
+			}
+
+			matcher := ConsistOf(
+				PointTo(MatchFields(IgnoreExtras, Fields{
+					"Type":     Equal(field.ErrorTypeInvalid),
+					"Field":    Equal("loggingRules.programNames[0]"),
+					"BadValue": Equal(""),
+					"Detail":   Equal(".programNames can't contain empty program name"),
+				})),
+			)
+
+			errorList := validation.ValidateRsyslogRelpConfig(&config, path)
+			Expect(errorList).To(matcher)
+		})
+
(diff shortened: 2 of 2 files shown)


pkg/apis/rsyslog/validation/rsyslog_relp.go
```diff
@@ -112,6 +121,17 @@ func validateLoggingRules(loggingRules []rsyslog.LoggingRule, fldPath *field.Pat
 	return allErrs
 }
 
+func validateProgramNames(programNames []string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	validatingRegex, _ := regexp.CompilePOSIX(`[[:/]`)
```
plkokanov: Can you move this outside the function (right before it on line 123 in a `MustCompile`
```
var programNamesRegex = regexp.MustCompille(`[[:/]`)

func validateProgramNames(programNames []string, fldPath *field.Path) field.ErrorList {
```

We do not need a POSIX regex here.

pkg/apis/rsyslog/validation/rsyslog_relp.go
```diff
@@ -112,6 +121,17 @@ func validateLoggingRules(loggingRules []rsyslog.LoggingRule, fldPath *field.Pat
...
+	for index, name := range programNames {
+		if validatingRegex.MatchString(name) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), name, ".programNames can't contain `[`, `:` or `/`"))
+		}
+	}
+	return allErrs
+}
```
plkokanov: This does not follow the requirement from the validation issuue: `checks if programNames contains only printable ASCI characters`

So maybe you could do something like this: 
```suggestion
var printableCharactersRegex = regexp.MustCompile(`^[!-~]*$`) 
var invalidCharactersForProgramNameRegex = regexp.MustCompile(`[[:/]`) 

func validateProgramNames(programNames []string, fldPath *field.Path) field.ErrorList {
	allErrs := field.ErrorList{}
	for index, name := range programNames {
		if  invalidCharactersForProgramNameRegex.MatchString(name) {
			allErrs = append(allErrs, field.Invalid(fldPath.Index(index), name, ".programNames can't contain `[`, `:` or `/`"))
		}
		if !printableCharactersRegex.MatchString(name) {
            allErrs = append(allErrs, field.Invalid(fldPath.Index(index), name, ".programName can only contain printable characters")
	}
	return allErrs
}

**Comment by RadaBDimitrova** [comment]
>General remark: why are you omitting the actual errors from the validation package functions and just checking their length? I think it's better if we include in the error message the reason for the error that comes from the validation function that we use.

It was mainly because `IsWildcardDNS1123Subdomain` and `IsDNS1123Subdomain` don't return `field.ErrorList`, but `[]string` with errors inside, which makes it more difficult to append. Also `MatchString` returns a bool, so I thought it may make sense to keep the errors uniform by doing this, but I see your vision here and I can adapt it.

File: pkg/apis/rsyslog/validation/rsyslog_relp.go [hunk 3]
```diff
@@ -112,6 +139,19 @@ func validateLoggingRules(loggingRules []rsyslog.LoggingRule, fldPath *field.Pat
 	return allErrs
 }
 
+func validateProgramNames(programNames []string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for index, name := range programNames {
+		if invalidCharactersForProgramNameRegex.MatchString(name) {
```
plkokanov: Sorry for forgetting to add this to the previous review. Could we also forbid program  names that are empty strings?

plkokanov (changes requested):
Just one small final comment on my side, that I missed in the previous review

